PR #10, opened 2021-09-10
Modify initialization routines to use new OpenFOAM data

MattRolchigo:
OpenFOAM no longer prints domain extent in temperature files - ParallelMeshInit has been modified to read temperature data twice: first to determine domain size and perform the decomposition, then to store the appropriate portions of the temperature data on the appropriate MPI ranks.

To account for the increased precision in OpenFOAM output, ExaCA reads the temperature data into a structure of type "double", rather than the "float" structure previously used

Removed functions from `CAInit` header that aren't needed downstream (for analysis)

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -9,6 +9,7 @@
 
 #include "mpi.h"
 
+#include <algorithm>
 #include <cmath>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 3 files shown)


src/runCA.cpp
```diff
@@ -60,9 +60,10 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
     int *FinishTimeStep = new int[NumberOfLayers];
 
     // Data structure for storing raw temperature data from file(s)
+    // Store data as double - needed for small time steps to resolve local differences in solidification conditions
     // With no remelting, each data point has 5 values (X, Y, Z coordinates, liquidus time, and either solidus time OR
     // cooling rate) Initial estimate for size
-    std::vector<float> RawData(1000000);
+    std::vector<double> RawData(1000000);
```
streeve, line 66: As usual, I'm suggesting we convert this to a View

src/CAinitialize.cpp
```diff
@@ -664,432 +664,344 @@ void ParallelMeshInit(int DecompositionStrategy, ViewI_H NeighborX, ViewI_H Neig
...
+                std::size_t ThirdComma = s.find(",", SecondComma + 1);
+                std::string XCoordinateS = s.substr(0, FirstComma);
+                std::string YCoordinateS = s.substr(FirstComma + 1, SecondComma - FirstComma - 1);
+                std::string ZCoordinateS = s.substr(SecondComma + 1, ThirdComma - SecondComma - 1);
+                XCoordinates[XYZPointCounter] = stod(XCoordinateS);
+                YCoordinates[XYZPointCounter] = stod(YCoordinateS);
+                ZCoordinates[XYZPointCounter] = stod(ZCoordinateS);
+                XYZPointCounter++;
+                if (XYZPointCounter == XCoordinates.size()) {
+                    XCoordinates.resize(XYZPointCounter + 1000000);
```
streeve, line 792: Why do you keep the extra buffer after counting?

src/CAinitialize.cpp
```diff
@@ -1286,7 +1198,7 @@ void TempInit_Reduced(int id, int &MyXSlices, int &MyYSlices, int &MyXOffset, in
         LayerID(i) = -1;
     }
 
-    float LayerwiseTSOffset = 0;
+    // removed unused LayerwiseTSOffset variable
```
streeve, line 1201: Is this important to know here?

streeve (commented):
This would be a good case to create a unit test for - correctly read in a very small FOAM data file

src/runCA.cpp
MattRolchigo: I feel like at one point there was a reason this structure was a vector instead of a view... but I don't see a reason it can't be a view now

src/CAinitialize.cpp
MattRolchigo: These buffers (XCoordinates, YCoordinates, ZCoordinates) have an initial guess for size (1000000), and are resized once they fills up (by adding 1000000 to the size). When the reading of points is complete, it is resized to the actual number of data points

src/CAinitialize.cpp
MattRolchigo: Kind of... this variable was just sitting there doing nothing, the new variable SmallestTime_Global takes on the role of what LayerwiseTSOffset was (at some point intended to do), aka make sure that each layer's "critical time step" starts at 0/time step 0 corresponds to the smallest value of "liquidus time" from the temperature data for the layer

src/runCA.cpp
MattRolchigo: I take that back, I played around with this and something about declaring RawData in RunProgram_Reduced, resizing RawData inside of ParallelMeshInit, and then passing it to TempInit_Reduced seems to mess things up and give an error when accessing RawData in TempInit_Reduced. I'm going to leave this as a vector for now, since the point of addressing this vector at all was to change the data type to account for added precisions from OpenFOAM, and converting this to a view may require more thought/debugging time than I expected

src/CAinitialize.cpp
```diff
@@ -664,432 +664,344 @@ void ParallelMeshInit(int DecompositionStrategy, ViewI_H NeighborX, ViewI_H Neig
...
+            if (col1 == std::string::npos)
+                col1 = HeaderLine.find("ts");
+            std::size_t col2 = HeaderLine.find("r");
+            if (col2 == std::string::npos)
+                col2 = HeaderLine.find("R");
+            if ((xf == std::string::npos) || (yf == std::string::npos) || (zf == std::string::npos) ||
+                (col1 == std::string::npos) || (col2 == std::string::npos)) {
+                std::string error = "One of the six required column headers did not appear in the "
+                                    "temperature input file: these are x, y, z, tm, tl, cr";
+                throw std::runtime_error(error);
```
streeve, line 762: Should create a small function for this and call for each variable - then the error can be more specific

src/CAinitialize.cpp
```diff
@@ -664,432 +664,344 @@ void ParallelMeshInit(int DecompositionStrategy, ViewI_H NeighborX, ViewI_H Neig
...
+            getline(TemperatureFile, DummyLine);
+
+            // Read data from the remaining lines - if temperature files have the old format (data separated by
+            // arbitrary number of spaces)
+            while (!TemperatureFile.eof()) {
+                std::string s;
+                getline(TemperatureFile, s);
+                if (s.empty())
+                    break;
+                std::size_t FirstComma = s.find(",");
```
streeve, line 943: Now that you do this twice, this should be a separate function passing the input line and line values

streeve (changes requested):
With a couple new functions to clean things up (and maybe there are others you can think of) this should be ready

src/CAinitialize.cpp
```diff
@@ -132,36 +131,43 @@ bool parseInputBool(std::ifstream &stream, std::string key) {
...
-                                     int LineNumber, std::string TemperatureFilename) {
-
-    if ((InputValue < LowerBound) || (InputValue > UpperBound)) {
-        std::string error = Label + " value " + std::to_string(InputValue) + " on line " + std::to_string(LineNumber) +
-                            " of file " + TemperatureFilename +
-                            " is invalid based on domain size limit given in the file header: must be greater than " +
-                            std::to_string(LowerBound) + " and less than " + std::to_string(UpperBound);
-        throw std::runtime_error(error);
+// Check to make sure that all expected column names appear in the header for this temperature file
+void checkForValues(std::string HeaderLine, std::string Check1, std::string Check2, std::string Check3) {
```
streeve, line 135: I like this, but this doesn't enforce the order assumed in the next function

MattRolchigo: Yeah, I realized that - what I had before didn't enforce that order either actually. I kind of want to change this section so that the header must have x, y, z, tm, tl, cr in that order, and any wrong capitalization or order would throw an error like:
"Header line must list, in order, x, y, z, tm ,tl , cr: the header line of the temperature data file instead lists [insert first line from file]"

streeve: You could allow "x y z tm tl cr" or "X Y Z Tm Tl Cr" and maybe all caps and nothing else

MattRolchigo: I think I'll change the function to take in the full header line, compare each of the 6 substrings by position against the expected values and maybe 2 alternative values, and then throw a position-specific runtime error if there's no match

streeve (changes requested):
Couple more changes

streeve (approved):

PR #14, opened 2021-09-16
Make temperature path configurable

streeve:
If a reasonable idea, we would need to update all the example input files and maybe do the same for substrates

### DIFF
--- a/examples/Inp_AMBenchMultilayer.txt
+++ b/examples/Inp_AMBenchMultilayer.txt
@@ -1,4 +1,4 @@
-Test problem using alternating odd-even layers of a cross-hatching scan pattern (with large CA cell size to keep the problem tractable): 
+Test problem using alternating odd-even layers of a cross-hatching scan pattern (with large CA cell size to keep the problem tractable):
 *****
 Problem type: R
 Decomposition strategy (1 for 1D domain decomposition, 2 for simple 2D decomposition, 3 for complex 2D decomposition): 1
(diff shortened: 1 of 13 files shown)


examples/Inp_SimpleRaster.txt
```diff
@@ -13,6 +13,7 @@ File of grain orientations:GrainOrientationVectors_Robert.csv
 Heat transport data mesh size (microns): 5
 Time step (microseconds): 0.08333333
 Substrate grain spacing (in microns): 25
+Path to temperature file(s): examples/Temperatures
 Temperature filename: Master-Raster-150um.txt
```
streeve, line 19: Add the same note as the AMbench file?

examples/Inp_AMBenchMultilayer.txt
```diff
@@ -2,18 +2,21 @@ Test problem using alternating odd-even layers of a cross-hatching scan pattern
 *****
 Problem type: R
 Decomposition strategy (1 for 1D domain decomposition, 2 for simple 2D decomposition, 3 for complex 2D decomposition): 1
+Path to material file: Materials
```
streeve, line 5: Will it ever make sense to have more than one file? We could switch the existing input from file name to full path (except temperature, unless there's a good way to do the file glob with c++)

streeve:
@MattRolchigo good news, I got this to work locally; bad news, I added a bug somewhere and the runs never finish (stuck at constant liquid cells)

MattRolchigo:
@streeve most likely that's a file path issue - I don't think I have guards on the grain orientation or material file for the case if they can't be opened, so ExaCA thinks it read A/B/C/D/GrainOrientationVectors, which is leading to non-physical grain orientations/growth kinetics

streeve:
> @streeve most likely that's a file path issue - I don't think I have guards on the grain orientation or material file for the case if they can't be opened, so ExaCA thinks it read A/B/C/D/GrainOrientationVectors, which is leading to non-physical grain orientations/growth kinetics

There are checks now for each file open in the last few commits. I'll take another look when I can

MattRolchigo (approved):

mikejwg:
Could you share the eternal data in the temperature_path for testing the Example problems only with external data. By the way, when I run the DirSolidification.txt,  some strange grain orientations were found in the solidification front.

streeve:
@mikejwg please open a new issue for this - this pull request has been merged and closed.

Issue #2, opened 2021-07-21
Unit tests


streeve:
Initial set of unit testing:

- [x] Test infrastructure
- [x] Steering vector
- [x] Cell capture 
- [x]  Ghosts
- [x]  Initialization
- [x]  Nucleation
- [x]  Output

streeve:
Lots of tests added and improved in #77 - initialization with remelting, nucleation, steering vector with remelting

streeve:
Added cell capture in #94

PR #15, opened 2021-09-20
Add unit test infrastructure

streeve:
Let's plan to merge this after adding the temperature read/init example. Of course if there are any other quick tests to add that's great

First step for #2

### DIFF
--- a/unit_test/CMakeLists.txt
+++ b/unit_test/CMakeLists.txt
@@ -1,0 +1,122 @@
+##---------------------------------------------------------------------------##
+# Download and unpack googletest
+##---------------------------------------------------------------------------##
+set(GTEST_URL "https://github.com/google/googletest/archive/release-1.10.0.tar.gz"
+    CACHE STRING "URL for GTest tarball")
+mark_as_advanced(GTEST_URL)
+
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL ${GTEST_URL}
+  URL_MD5         ecd1fa65e7de707cd5c00bdac56022cd
+  )
+
+FetchContent_GetProperties(googletest)
+if(NOT googletest_POPULATED)
+  FetchContent_Populate(googletest)
+  add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR})
+endif()
+
+# Prevent GoogleTest from overriding our compiler/linker options
+# when building with Visual Studio
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+
+set(gtest_args --gtest_color=yes)
+
+##--------------------------------------------------------------------------##
+## Add backends
+##--------------------------------------------------------------------------##
+set(EXACA_TEST_DEVICES)
+foreach(_device ${EXACA_SUPPORTED_DEVICES})
+  if(Kokkos_ENABLE_${_device})
+    list(APPEND EXACA_TEST_DEVICES ${_device})
+  endif()
+endforeach()
+
+##--------------------------------------------------------------------------##
+## Create tests without Kokkos
+##--------------------------------------------------------------------------##
+macro(ExaCA_add_tests_nobackend)
+  cmake_parse_arguments(EXACA_UNIT_TEST "" "" "NAMES" ${ARGN})
+  set(EXACA_UNIT_TEST_MPIEXEC_NUMPROCS 1)
+  foreach( _np 2 4 )
+    if(MPIEXEC_MAX_NUMPROCS GREATER_EQUAL ${_np})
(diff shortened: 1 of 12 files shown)


MattRolchigo (approved):

streeve:
@MattRolchigo Right now we have a passing blank test - can you add to the `tstInit` file?

streeve (commented):
Let's add one more test that actually uses Kokkos, plus my comments below

unit_test/tstInit.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::ofstream TestDataFile;
+    TestDataFile.open(TestTempFileName);
+    TestDataFile << "x, y, z, tm, tl, cr" << std::endl;
+    for (int i = 0; i < 17; i++) {
+        TestDataFile << TestData[i][0] << "," << TestData[i][1] << "," << TestData[i][2] << "," << TestData[i][3] << ","
+                     << TestData[i][4] << "," << TestData[i][5] << std::endl;
+    }
+    TestDataFile.close();
+    // Read in data to "RawData"
+    ReadTemperatureData(deltax, HT_deltax, MyXSlices, MyYSlices, MyXOffset, MyYOffset, nx, ny, XMin, YMin, temp_paths,
```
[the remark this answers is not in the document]
MattRolchigo: It makes a "fake" decomposition, basically just assigning each rank one of four possible overlapping pieces of the grid. Since the routine involves reading only data relevant to each rank's known subdomain, I had to assign different ranks different subdomains to ensure that ranks aren't storing information they shouldn't be (or ignoring information they need)

unit_test/tstInit.hpp
```diff
@@ -0,0 +1,138 @@
...
+    MPI_Comm_rank(MPI_COMM_WORLD, &id);
+    int Rem = id % 4;
+    // Create test data.
+    int MyXOffset, MyYOffset;
+    double deltax = 1 * pow(10, -6);
+    double HT_deltax = 1 * pow(10, -6);
+    int MyXSlices = 8;
+    int MyYSlices = 14;
+    if (Rem == 0) {
+        MyXOffset = 0;
```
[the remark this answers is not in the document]
MattRolchigo: As long as the number of ranks is less than or equal to 4, that would work. Otherwise the other ranks would be doing no work while the first four did (granted, in this situation I'm assigning redundant work to the other ranks anyways)...

unit_test/tstInit.cpp
```diff
@@ -0,0 +1,124 @@
...
+    int NumberOfLayers = 1;
+    int TempFilesInSeries = 1;
+    unsigned int NumberOfTemperatureDataPoints = 0;
+    std::vector<double> RawData(12);
+    int *FirstValue = new int[NumberOfLayers];
+    int *LastValue = new int[NumberOfLayers];
+    int GlobalProcCols = (np + 1) / 2;
+    int AllRanks_NDataPoints = 6 * GlobalProcCols;
+    double TestData[AllRanks_NDataPoints][6];
+    for (int i = 0; i < GlobalProcCols; i++) {
```
streeve, line 51: Can you add some comments about what this data is meant to be? I like this new test a lot more, but I don't really get the intent

unit_test/CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+              ${_target} ${MPIEXEC_POSTFLAGS} ${gtest_args})
+        endif()
+      endforeach()
+    endforeach()
+  endforeach()
+endmacro()
+
+ExaCA_add_tests_nobackend(NAMES Init)
+
+#ExaCA_add_tests(NAMES)
```
streeve, line 122:
```suggestion
ExaCA_add_tests(NAMES SubstrateInit)
```

unit_test/tstSubstrateInitFGS.hpp
```diff
@@ -0,0 +1,145 @@
...
+        MyYOffset = 0;
+    else
+        MyYOffset = 2;
+    double deltax = 1 * pow(10, -6);
+    int LocalActiveDomainSize = MyXSlices * MyYSlices * nzActive;
+    int LocalDomainSize = MyXSlices * MyYSlices * nz;
+    // There are 48 total cells in this domain (nx * ny * nz)
+    // Even ranks have 30, and odd ranks have the other 30
+    double SubstrateGrainSpacing = 1.95; // This grain spacing ensures that there will be 3 substrate grains
+    Kokkos::View<int *, layout, TEST_MEMSPACE> GrainID_H(Kokkos::ViewAllocateWithoutInitializing("GrainID"),
```
streeve, line 42: I think using `CAtypes` is reasonable

streeve (commented):
Looks good, with some minor clean up

PR #16, opened 2021-09-22
Use of Kokkos::fence to avoid ghost nodes init race conditions

MattRolchigo:


streeve:
Can you explain the issue a little more? I wouldn't expect needing this without using UVM

MattRolchigo:
So what I believe was happening for certain problems (including the SpotMelt problem) and certain numbers of MPI ranks (with greater odds with more ranks) was that the buffers were being sent before their filling (inside the loop labeled "GNInit") was finished. Similarly, the unpacking of the buffers (inside the loop labeled "BufferUnpack") must be done before any of the new octahedron calculations (the second loop labeled "GNInit") - otherwise octahedron calculations may not be performed on all the required cells. Then I put a final "fence" at the end of the last parallel loop to ensure that all ghost node data was initialized before moving on with the code.

streeve:
If this clearly fixed a specific issue, I'll approve. I just wanted to make sure that was the case

MattRolchigo:
I just linked it to the associated issue (which I forgot that I made)

streeve (approved):

streeve:
Ready to merge @MattRolchigo

PR #20, opened 2021-10-04
Added input files for OpenFOAM and ExaConstit workflow

MattRolchigo:


streeve (approved):
I approve, but I didn't test - we can update as needed for the workflow runs

PR #29, opened 2021-10-20
Removed unneeded wall cells

MattRolchigo:
Removed wall cells at top (Z = nz-1) surface; ensure consistency with domain dimensions and definition of top/bottom surfaces across problem types, regardless of the number of layers.

Consistent (across problem types) definition of the location of the baseplate top (for spot melt simulations and for simulations using read temperature data) as the largest Z coordinate in the first layer containing temperature data


src/CAprint.cpp
```diff
@@ -19,7 +19,7 @@ void CollectIntField(ViewI3D_H IntVar_WholeDomain, ViewI_H IntVar, int nz, int M
                      ViewI_H RecvXOffset, ViewI_H RecvYOffset, ViewI_H RecvXSlices, ViewI_H RecvYSlices,
                      ViewI_H RBufSize) {
 
-    for (int k = 1; k < nz - 1; k++) {
+    for (int k = 0; k < nz; k++) {
```
streeve, line 22: So here you are printing the wall cells?

src/CAinitialize.cpp
```diff
@@ -1646,7 +1667,7 @@ void SubstrateInit_FromGrainSpacing(float SubstrateGrainSpacing, int nx, int ny,
     Kokkos::parallel_for(
         "CellCapture",
         Kokkos::MDRangePolicy<Kokkos::Rank<3, Kokkos::Iterate::Right, Kokkos::Iterate::Right>>(
-            {1, 0, 0}, {nzActive, MyXSlices, MyYSlices}),
+            {1, 0, 0}, {nzActive + 1, MyXSlices, MyYSlices}),
```
streeve, line 1670: Why the +1 here?

streeve (commented):
Is removing the top wall cells change the problem physically? Or is it only for visualization?

src/CAprint.cpp
MattRolchigo: Not necessarily - but all cells should be gathered and sent to rank 0, including the wall cells

src/CAinitialize.cpp
MattRolchigo: The cells at Z = nzActive were considered part of the powder previously - but since Z = 1 through Z = nzActive is now considered "layer 0", the powder layer should start at Z = nzActive + 1

MattRolchigo:
Removing the wall cells is for visualization, but also because they are not needed. There is already a check to ensure that no cells beyond the top surface of a given layer are accessed, which means that the wall cells at the top surface of the problem domain itself don't do anything

streeve (approved):

src/CAinitialize.cpp
MattRolchigo: Yes - I should probably clarify that in the PR description, since as of right now I only have a vague mention of substrate

PR #31, opened 2021-11-01
Init modularize

MattRolchigo:
Split ParallelMeshInit into more modular initialization routines


streeve (approved):
As far as I can tell, this is just slight rearrangement and new functions, right?

streeve:
Should squash rebase or squash merge

MattRolchigo:
Yes, this removes the function ParallelMeshInit and replaces it with 4 smaller functions with more intuitive names

PR #40, opened 2022-01-28
Use physical XYZ bounds for plotting vtk data

MattRolchigo:
Updated the printing subroutines PrintCAData and PrintGrainMisorientations for consistency with PrintIntermediateExaCAState - vtk files should associated CA data with physical XYZ coordinate values, rather than integer CA cell positions

### DIFF
--- a/src/CAprint.cpp
+++ b/src/CAprint.cpp
@@ -258,11 +258,12 @@ void PrintExaCAData(int id, int layernumber, int np, int nx, int ny, int nz, int
 
         if (PrintMisorientation)
             PrintGrainMisorientations(BaseFileName, PathToOutput, nx, ny, nz, Melted_WholeDomain, GrainID_WholeDomain,
-                                      GrainOrientation, GrainUnitVector, NGrainOrientations);
+                                      GrainOrientation, GrainUnitVector, NGrainOrientations, deltax, XMin, YMin, ZMin);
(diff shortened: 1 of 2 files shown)


src/CAprint.cpp
```diff
@@ -421,8 +423,8 @@ void PrintGrainMisorientations(std::string BaseFileName, std::string PathToOutpu
     GrainplotM << "ASCII" << std::endl;
     GrainplotM << "DATASET STRUCTURED_POINTS" << std::endl;
     GrainplotM << "DIMENSIONS " << nx - 2 << " " << ny - 2 << " " << nz - 1 << std::endl;
-    GrainplotM << "ORIGIN 0 0 0" << std::endl;
```
streeve, line 424: I feel like I probably should have questioned this when you added it..

src/CAprint.cpp
```diff
@@ -421,8 +423,8 @@ void PrintGrainMisorientations(std::string BaseFileName, std::string PathToOutpu
...
-    GrainplotM << "SPACING 1 1 1" << std::endl;
+    GrainplotM << "ORIGIN " << XMin - deltax << " " << YMin - deltax << " " << ZMin - deltax << std::endl;
```
streeve, line 426: Why do you subtract `deltax` from the min values?

streeve (approved):
Looks good, but I don't understand the one point

PR #42, opened 2022-02-08
Made timers for init/calc/print consistent

MattRolchigo:
Corrects bug in calculation of "time spent performing CA calculations", calculates time spent initializing data, performing CA calculations, and exporting data in a consistent manner


streeve:
I don't see what was inconsistent - does this fix what you saw with HIP?

MattRolchigo:
The way that "InitTime" was calculated was different than the ways that RunTime and OutTime were calculated - this change set makes sure that a call to the the timer is performed at the start of the section of the code undergoing timing, and a call to the timer is performed at the end, with the total time for a section not depending on other values. Previously, this inconsistency was the thing leading to erroneous time calculations with the HIP backend

streeve (approved):

PR #43, opened 2022-02-16
Moved file parsing functions into a new file

MattRolchigo:
Functions from src/CAinitialize.cpp and analysis/src/GAutils.cpp that were involved with parsing lines of input files were moved into a new file, src/CAparsefiles.cpp


streeve (approved):
I would just call it CAparse, but either works for me

PR #44, opened 2022-02-18
Separate general/problem-specific inputs during input file read

MattRolchigo:
Use separate vectors for general and problem type-specific input strings and values read from files. This will make it easier to add new general required/optional inputs in the future without having to renumber the others

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -38,7 +38,7 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
     RemeltingYN = false;
 
     // Required inputs that should be present in the input file, regardless of problem type
-    std::vector<std::string> RequiredInputs = {
+    std::vector<std::string> RequiredInputs_General = {
         "Decomposition strategy",                        // Required input 0
         "Material",                                      // Required input 1
         "Cell size",                                     // Required input 2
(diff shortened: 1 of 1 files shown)


src/CAinitialize.cpp
```diff
@@ -76,72 +76,82 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
...
+            // Check if this is a general required input
+            bool RequiredYN =
+                parseInputFromList(line, RequiredInputs_General, RequiredInputsRead_General, NumRequiredInputs_General);
             if (!(RequiredYN)) {
-                bool OptionalYN = parseInputFromList(line, OptionalInputs, OptionalInputsRead, NumOptionalInputs);
+                // If not a general required input, check if this is a problem type specific required input
+                RequiredYN = parseInputFromList(line, RequiredInputs_ProblemSpecific,
+                                                RequiredInputsRead_ProblemSpecific, NumRequiredInputs_ProblemSpecific);
+            }
+            if (!(RequiredYN)) {
```
streeve, line 145: Can you use separate bools for these? I get confused what this logic is doing

streeve (approved):
Looks good - not a bad time for a super simple input value unit test to guard against dumb mistakes

src/CAinitialize.cpp
streeve: Oops never mind

PR #46, opened 2022-02-18
Calculation of grid size ratio, deltax adjustment moved

MattRolchigo:
The adjustment of deltax to ensure that the ratio of temperature data and CA data grid sizes (HTtoCAratio) is calculated when the quantities are first used in the reading of temperature data. HTtoCAratio is calculated in ReadTemperatureData (modifying the ReadTemperatureData unit test to accommodate this), and used again in TempInit_Reduced, rather than recalculated

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -730,12 +730,26 @@ void DomainDecomposition(int DecompositionStrategy, int id, int np, int &MyXSlic
 }
 
 // Read in temperature data from files, stored in "RawData", with the appropriate MPI ranks storing the appropriate data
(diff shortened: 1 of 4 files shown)


streeve (commented):
You mention that you previously recalculated the ratio, but I only see it moved from one place to another

MattRolchigo:
It was previously calculated on line 738 and 965 in CAinitialize.cpp, now it is an argument passed to both init subroutines, calculated in ReadTemperatureData, and reused in TempInit_Reduced

streeve (approved):

PR #47, opened 2022-02-21
Initialize send/recv buffers to 0

MattRolchigo:
The send and receive buffers, after allocation, should be initialized to 0. Currently, their values are accessed prior to the placement of any data in the ghost nodes, which will sometimes cause a segmentation fault when running with MPI

### DIFF
--- a/src/runCA.cpp
+++ b/src/runCA.cpp
@@ -221,23 +221,23 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
         BufSizeY = MyYSlices - 2;
         BufSizeZ = nzActive;
     }
-    Buffer2D BufferSouthSend(Kokkos::ViewAllocateWithoutInitializing("BufferSouthSend"), BufSizeX * BufSizeZ, 5);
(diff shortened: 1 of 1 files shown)


src/runCA.cpp
```diff
@@ -237,7 +237,25 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
     Buffer2D BufferNorthWestRecv(Kokkos::ViewAllocateWithoutInitializing("BufferNorthWestRecv"), BufSizeZ, 5);
     Buffer2D BufferSouthEastRecv(Kokkos::ViewAllocateWithoutInitializing("BufferSouthEastRecv"), BufSizeZ, 5);
     Buffer2D BufferSouthWestRecv(Kokkos::ViewAllocateWithoutInitializing("BufferSouthWestRecv"), BufSizeZ, 5);
```
streeve, line 239: Instead, remove the `Kokkos::ViewAllocateWithoutInitializing` above, add a comment saying that these do need zero init (so that we don't try to optimize later), and remove the deep_copies

streeve (approved):
Need to rebase on master

PR #48, opened 2022-02-22
Ensure activation of cells at bottom boundary during init

MattRolchigo:
Previous criteria of liquid cells that border at least one solid cell during initialization transforming into active cells is insufficient, as liquid cells at the domain's bottom surface technically border solid material despite potentially only bordering other liquid cells (the region below these liquid cells would be solid, if the simulation domain extended into the baseplate). This is corrected by adding a second criteria for liquid to active cell transformation during ActiveCellInit in src/CAinitialize.cpp


streeve (approved):
Need to rebase on master to remove the previous commit

PR #49, opened 2022-02-22
Fix access of 0D view data

MattRolchigo:
Access 0D view data using "Data(0)" rather than "Data()", to avoid Kokkos error

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -1426,7 +1426,7 @@ void SubstrateInit_FromGrainSpacing(float SubstrateGrainSpacing, int nx, int ny,
             // (grain ID = 0 is not used)
             float MinDistanceToThisGrain = (float)(LocalActiveDomainSize);
             int ClosestGrainIndex = -1;
-            for (int n = 0; n < NumBaseplateGrains_G(); n++) {
+            for (int n = 0; n < NumBaseplateGrains_G(0); n++) {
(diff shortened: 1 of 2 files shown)


src/CAinitialize.cpp
```diff
@@ -1534,20 +1534,29 @@ void ActiveCellInit(int id, int MyXSlices, int MyYSlices, int nz, ViewI_H CellTy
...
-                            int NeighborD3D1ConvPosition =
-                                MyNeighborZ * MyXSlices * MyYSlices + MyNeighborX * MyYSlices + MyNeighborY;
-                            if ((MyNeighborX >= 0) && (MyNeighborX < MyXSlices) && (MyNeighborY >= 0) &&
-                                (MyNeighborY < MyYSlices) && (MyNeighborZ >= 0) && (MyNeighborZ < nz)) {
-                                if (CellType(NeighborD3D1ConvPosition) == Solid) {
-                                    // This cell is at the interface - becomes active type
-                                    CellType(CAGridLocation) = Active;
-                                    SubstrateActCells_ThisRank++;
-                                    return;
+                        if (k == 0) {
```
streeve, line 1537: This seems unrelated

MattRolchigo: That is unrelated, that change is the subject of a different PR. I must have messed up when rebasing

streeve (approved):

PR #52, opened 2022-03-15
Added option to print final cell undercooling values to file

MattRolchigo:


src/runCA.cpp
```diff
@@ -456,16 +457,19 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
     // Copy GPU results for GrainID back to CPU for printing to file(s)
     Kokkos::deep_copy(GrainID_H, GrainID_G);
     Kokkos::deep_copy(CellType_H, CellType_G);
+    // Copy final undercooling values back to the CPU for potential printing to file
+    Kokkos::deep_copy(UndercoolingCurrent_H, UndercoolingCurrent_G);
```
streeve, line 461: Can't you wrap this in an if block if you're not outputting?

streeve (approved):

PR #53, opened 2022-03-17
Added default RVE printing option to ExaCA

MattRolchigo:
Optional input for test problems using input time-temperature history data to print a "default" (center of domain in X and Y, closest position to the top surface in Z while avoiding last layer's microstructure) 0.5 cubic mm representative volume element (RVE) used by ExaConstit as part of the main ExaCA code. The ability to print "custom" RVE data (various sizes and locations in the domain) still requires running the separate analysis executable, which reads ExaCA output data and parses it.

Part of #2 (output)


streeve (changes requested):
Conflict needs to be fixed

src/CAinitialize.cpp
```diff
@@ -348,6 +350,12 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
             std::cout << "Note: optional input ExtraWalls is no longer used, all simulations by default have no walls "
                          "cells along domain boundaries"
                       << std::endl;
+        // Should optional RVE data be printed for the standard location (center of domain in X and Y, as close to the
+        // top of the domain in Z as possiblewithout including the last layer's microstructure)?
+        if (OptionalInputsRead_ProblemSpecific[5].empty())
```
streeve, line 355: Make it false outside the if block and check if it's not empty

streeve (approved):

PR #55, opened 2022-03-18
Removed unneeded ItList, GrainOrientation views

MattRolchigo:
ItList previous was used to control which neighboring cells should be iterated over, but the extra memory access of checking ItList instead of checking all 26 neighbors of a cell and determining which neighbors were in bounds was slower and required more lines of code. GrainOrientation was used to map GrainID values to grain orientations; however, since the code uses the Grain ID values themselves to correspond with grain orientations, the GrainOrientation view was redundant. On all backends, performance was either unchanged or very slightly improved (<1% of total code run time)


src/CAupdate.cpp
```diff
@@ -64,6 +64,12 @@ struct reduction_identity<sample::ValueType> {
 };
 } // namespace Kokkos
 
+// Get the orientation of a grain from a given grain ID and the number of possible orientations
+KOKKOS_INLINE_FUNCTION int getGrainOrientation(int MyGrainID, int NGrainOrientations) {
```
streeve, line 68: Since everything seems to use it, I guess it should be in CAfunctions

streeve (approved):

PR #72, opened 2022-05-30
Use double (not float) for Buffer2D type; removed unused Buffer3D type

MattRolchigo:
For large values of GrainID (integers) that are often encountered in large multilayer problems, the conversion to float and back to int during the loading, sending, and unpacking of halo data was leading to values being erroneously altered. Redefining the Buffer2D type to use double values instead of float values mitigates this issue.

### DIFF
--- a/src/CAupdate.cpp
+++ b/src/CAupdate.cpp
@@ -441,11 +441,11 @@ void CellCapture(int np, int cycle, int DecompositionStrategy, int LocalActiveDo
 
                                 if (np > 1) {
 
-                                    float GhostGID = h;
-                                    float GhostDOCX = cx;
-                                    float GhostDOCY = cy;
-                                    float GhostDOCZ = cz;
-                                    float GhostDL = NewODiagL;
+                                    double GhostGID = (double)(h);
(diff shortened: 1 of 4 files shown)


src/CAupdate.cpp
```diff
@@ -441,11 +441,11 @@ void CellCapture(int np, int cycle, int DecompositionStrategy, int LocalActiveDo
 
                                 if (np > 1) {
 
-                                    float GhostGID = h;
-                                    float GhostDOCX = cx;
-                                    float GhostDOCY = cy;
-                                    float GhostDOCZ = cz;
-                                    float GhostDL = NewODiagL;
+                                    double GhostGID = (double)(h);
```
streeve, line 444: So you're doing this cast to avoid a separate int buffer?

MattRolchigo: Yes, I wanted to avoid an additional send/recv operation. Theoretically, if we had a 2D view of size N by 5 created with a custom type where an N by 1 int view was combined with an N by 4 float view, we could avoid these casts as well

PR #74, opened 2022-06-02
Temporary host view size calc/initialization moved out of main/runCA.cpp

MattRolchigo:
Created new functions or modified existing subroutines to allocate host copies of device views only in the scope needed


src/CAinitialize.cpp
```diff
@@ -396,93 +396,103 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
 
 //*****************************************************************************/
 // Intialize neighbor list structures (NeighborX, NeighborY, NeighborZ)
-void NeighborListInit(ViewI_H NeighborX, ViewI_H NeighborY, ViewI_H NeighborZ) {
+void NeighborListInit(ViewI &NeighborX, ViewI &NeighborY, ViewI &NeighborZ) {
+
+    // Temporary host views for initialization
+    ViewI_H NeighborX_Host(Kokkos::ViewAllocateWithoutInitializing("NeighborX_Host"), 26);
```
streeve, line 402: This will be much simpler as a fixed-size array: `Kokkos::Array<int, 26> NeighborX = {0, -1, 0, ...};`

src/CAinitialize.cpp
```diff
@@ -795,63 +805,129 @@ void ReadTemperatureData(int id, double &deltax, double HT_deltax, int &HTtoCAra
...
-    // Contrained solidification test problem
-    ZBound_Low = 0;
-    ZBound_High = nz - 1;
-    nzActive = ZBound_High - ZBound_Low + 1;
+// Get the Z coordinate of the lower bound of iteration
+// If using remelting, this is called before initializing temperature data views for each layer
+// If not using remelting, this is only called before initializing temperature data view for the first layer, since the
+// existing DomainShiftAndResize subroutine already exists for other layers. These functions should be unified in a
+// future update
+int calcZBound_Low(bool RemeltingYN, std::string SimulationType, int LayerHeight, int layernumber, float *ZMinLayer,
```
streeve, line 818: Shouldn't this be in the next PR?

src/CAinitialize.cpp
```diff
@@ -888,39 +964,37 @@ void TempInit_SpotMelt(double G, double R, std::string, int id, int &MyXSlices,
...
-        for (int i = 0; i < MyXSlices; i++) {
-            for (int j = 0; j < MyYSlices; j++) {
-                int Coord3D1D = k * MyXSlices * MyYSlices + i * MyYSlices + j;
-                CritTimeStep(Coord3D1D) = 0;
-                UndercoolingChange(Coord3D1D) = 0.0;
-                UndercoolingCurrent(Coord3D1D) = 0.0;
-            }
-        }
-    }
+    // These views are initialized to zeros on the host, filled with data, and then copied to the device for layer
```
streeve, line 998: Zero init because the full domain isn't filled below?

src/CAinitialize.cpp
```diff
@@ -795,63 +810,129 @@ void ReadTemperatureData(int id, double &deltax, double HT_deltax, int &HTtoCAra
...
-                CritTimeStep(GlobalD3D1ConvPosition) = (int)((k * G * deltax) / (R * deltat));
+                UndercoolingChange_Host(GlobalD3D1ConvPosition) = R * deltat;
+                CritTimeStep_Host(GlobalD3D1ConvPosition) = (int)((k * G * deltax) / (R * deltat));
                 Melted[GlobalD3D1ConvPosition] = true;
             }
         }
     }
+
+    // Copy initialized host data back to device
+    CritTimeStep = Kokkos::create_mirror_view_and_copy(Kokkos::DefaultExecutionSpace(), CritTimeStep_Host);
```
streeve, line 910: Let's typedef the `DefaultExecutionSpace` and use that. Should make it easier to run on non-default if we want to

streeve (changes requested):

src/CAinitialize.cpp
MattRolchigo: For CritTimeStep and UndercoolingChange, yes, because cells that don't have associated temperature data should have zeros for these. For LayerID, the initial value for all cells is -1, so I can change that view to allocate without initializing

src/runCA.cpp
```diff
@@ -53,10 +53,12 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
         NeighborRank_NorthWest, NeighborRank_SouthEast, NeighborRank_SouthWest;
     // Variables denoting whether or not each MPI rank's grid is at a global domain boundary
     bool AtNorthBoundary, AtSouthBoundary, AtEastBoundary, AtWestBoundary;
+    // View initialization is performed on host views, but copied to the device views within subroutines
+    using memory_space = Kokkos::DefaultExecutionSpace::memory_space;
```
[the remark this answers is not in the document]
streeve: Add this to CAtypes and you can use it everywhere you currently explicitly use the default in a mirror copy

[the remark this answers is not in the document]
MattRolchigo: So two lines in CAtypes that look like:

using device_memory_space = Kokkos::DefaultExecutionSpace::memory_space;
using host_memory_space = Kokkos::HostSpace::memory_space;

and then using those throughout in the mirror copy statements?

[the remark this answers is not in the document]
streeve: I would just use HostSpace directly, but the first one yes

src/CAghostnodes.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include "CAtypes.hpp"
 
+#include <Kokkos_Array.hpp>
```
streeve, line 11: This is in Core

src/CAinitialize.cpp
streeve: This line still needs to be removed

src/CAinitialize.cpp
streeve: Makes sense. Still kind of borderline for inclusion

src/CAinitialize.cpp
```diff
@@ -795,63 +727,129 @@ void ReadTemperatureData(int id, double &deltax, double HT_deltax, int &HTtoCAra
...
+        ZBound_High = SpotRadius + LayerHeight * layernumber;
+    }
+    else if (SimulationType == "R") {
+        // Top of layer comes from the layer's file data (implicitly assumes bottom of layer 0 is the bottom of the
+        // overall domain - this should be fixed in the future for edge cases where this isn't true)
+        ZBound_High = round((ZMaxLayer[layernumber] - ZMin) / deltax);
+    }
+    if (ZBound_High == -1)
+        throw std::runtime_error(
+            "Error: ZBound_High went uninitialized, problem type may not have been a valid option");
```
streeve, line 782: Why is this different than the low bound?

src/CAinitialize.cpp
streeve: Unchanged?

streeve (commented):
Should be ready after this

src/CAinitialize.cpp
streeve: These should use `memory_space()`, shouldn't they?

src/CAinitialize.cpp
streeve: Ok. Adding "must be S or R" is still helpful

streeve (approved):

PR #79, opened 2022-06-17
Update grain orientation post-processing print routines

MattRolchigo:
The file of Euler angles was updated, and OrientationInit was customized to read either the file of rotation matrix data or the file of Euler angle data. The post-processing routines for plotting inverse pole figure-coloring of microstructure and pole figures were updated to print the data in the form needed by Matlab/MTEX, which eliminates an extra step from the process of plotting the data


analysis/src/runGA.cpp
```diff
@@ -21,15 +24,19 @@
 int main(int argc, char *argv[]) {
 
     // Read command line input to obtain name of analysis file
-    std::string AnalysisFile, LogFile, MicrostructureFile, RotationFilename, OutputFileName;
+    bool NewOrientationFormatYN;
```
streeve, line 27: Isn't it better to store what the format is and check that?

src/runCA.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "CAghostnodes.hpp"
 #include "CAinitialize.hpp"
+#include "CAparsefiles.hpp"
```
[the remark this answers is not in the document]
MattRolchigo: CAinitialize and CAparsefiles are both quickly becoming very bloated, I think we should have a separate PR deciding how to reorganize the subroutines/rename these files

src/CAinitialize.hpp
```diff
@@ -57,7 +57,8 @@ void TempInit_Reduced(int id, int &MyXSlices, int &MyYSlices, int &MyXOffset, in
                       ViewF &UndercoolingChange, float XMin, float YMin, float ZMin, bool *Melted, float *ZMinLayer,
                       float *ZMaxLayer, int LayerHeight, int NumberOfLayers, int *FinishTimeStep, double FreezingRange,
                       ViewI &LayerID, int *FirstValue, int *LastValue, std::vector<double> RawData);
-void OrientationInit(int id, int NGrainOrientations, ViewF &GrainUnitVector, std::string GrainOrientationFile);
+void OrientationInit(int id, int &NGrainOrientations, ViewF &ReadOrientationData, std::string GrainOrientationFile,
+                     int ValsPerLine = 9);
```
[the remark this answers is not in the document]
MattRolchigo: The "default" behavior is to read the rotation matrix components from "GrainOrientationVectors_Robert.csv", of which there are 9 per orientation - which let me avoid touching the existing calls to OrientationInit. The "3" can be passed to use the same subroutine for reading the file of euler angles, since there's 3 per orientation

unit_test/tstKokkosInit.hpp
```diff
@@ -14,6 +15,56 @@
...
 //---------------------------------------------------------------------------//
+void testOrientationInit() {
+
+    int id;
+    // Get individual process ID
+    MPI_Comm_rank(MPI_COMM_WORLD, &id);
+
+    std::string GrainOrientationFile;
+    int ValsPerLine, NGrainOrientations = 0;
+    // Even ranks read unit vectors file, odd ranks read euler angles file
```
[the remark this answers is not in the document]
streeve: It will have MPI initialized, but there's no reason to run it with multiple ranks. I think it's a lot more clear to run the same test twice than to use different ranks for different cases. Plus that leaves half untested when only using one rank

analysis/src/runGA.cpp
MattRolchigo: I mean, I guess I could rename it "ReadEulerAngles" or something like that, but that's also not intuitive on how it affects the format of the orientation data that gets printed, since it's not as simple as swapping out one type of orientation for the other. The convention new/old only appears in this one variable, and it's hopefully something we can remove soon when we remove the option of allowing the "old" format

analysis/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Section 1: ExaConstit representative volume element (RVE) printing
+
+Representative volume elements can be carved out from the ExaCA results and used as input to the crystal plasticity FEM code ExaConstit (https://github.com/LLNL/ExaConstit). Lines in this section have the format [size, center x, center y, center z] - multiple RVEs can be printed as long as they are formatted as above and on separate lines.
+* Size of the RVE domain is given in cells (must be less than domain dimensions)
+* Center x, center y, and center z are the location (in cells) of the center of this RVE (If "D" is given as any of the inputs, the RVE will be taken to be as close to the domain center in x and y as possible, and as close to the top of the domain in z as possible while avoiding the last layer's microstructure)
+* If no ExaConstit data should be printed, no lines should exist between the short and long lines of asterisks here
+
+## Section 2: Data for plotting microstructure using MTEX
```
streeve, line 30: Only MTEX for now?

src/CAparsefiles.cpp
```diff
@@ -375,3 +375,28 @@ std::string parseCoordinatePair(std::string line, int val) {
     SplitStr = std::regex_replace(SplitStr, r, "");
     return SplitStr;
 }
+
+// Separate ReadLine at commas and place components into ParsedLine, checking to make sure that the size of ParsedLine
+// is equal to the number of comma-separated values in ReadLine
+void parseCommaSeparatedArgs(std::string AnalysisFile, std::string ReadLine, std::vector<std::string> &ParsedLine) {
```
streeve, line 381: Should this be used in temperature reading in the next PR?

streeve (approved):
Looks good. Ready to merge after the small cleanup suggested

analysis/README.md
MattRolchigo: While there might be a way to alter the format to make sure it can be used by MTEX or pyEBSD, I've only tested it with MTEX to this point

src/CAparsefiles.cpp
MattRolchigo: Both "ReadTemperatureData" and "FindXYZBounds" should use the new "parseCommaSeparatedArgs", as opposed to what they have now (which is basically the same behavior but as a combination of other lines and an older function "getTemperatureDataPoint"). I think that should be a separate PR from both this PR and the remeltlogic PR (behavior of ReadTemperatureData/FindXYZBounds is the same with/without remelting

analysis/src/GAutils.cpp
```diff
@@ -153,13 +163,29 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string
...
     RotationFilename = checkFileInstalled(RotationFilename, "Substrate", 0);
     // Check that files are not empty
     checkFileNotEmpty(LogFile);
     checkFileNotEmpty(MicrostructureFile);
     checkFileNotEmpty(RotationFilename);
+    // Same checks for EulerAnglesFilename, if given
+    if (NewOrientationFormatYN) {
+        EulerAnglesFilename = checkFileInstalled(EulerAnglesFilename, "Substrate", 0);
+        checkFileNotEmpty(EulerAnglesFilename);
+        std::cout << EulerAnglesFilename << std::endl;
```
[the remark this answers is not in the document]
MattRolchigo: Nope

PR #82, opened 2022-06-24
Use host view for histogram of orientation in analysis script

MattRolchigo:
Fixup of #79


streeve (approved):
#59 would have caught this...

PR #84, opened 2022-07-28
Replaced parseCommaSeparatedArgs function usage

MattRolchigo:
Replaced usage of the (slow) parseCommaSeparatedArgs function by expanding the functionality of the splitString function (fast).

Fixup of #81

### DIFF
--- a/src/CAparsefiles.cpp
+++ b/src/CAparsefiles.cpp
@@ -123,18 +123,13 @@ double getInputDouble(std::string val_input, int factor = 0) {
 
 // Given a string ("line"), parse at "separator" (commas used by default)
 // Modifies "parsed_line" to hold the comma separated values
-void splitString(std::string line, std::vector<std::string> &parsed_line, std::string separator = ",",
-                 int expected_num_commas = -1) {
+// expected_num_commas may be larger than the size of parsed_line, if only a portion of the line is being parsed
+void splitString(std::string line, std::vector<std::string> &parsed_line, int expected_num_commas,
+                 char separator = ',') {
     std::size_t line_size = parsed_line.size();
     // If reading a comma-separated line, make sure the right number of commas are present
-    if (separator == ",") {
-        int num_commas = std::count(line.begin(), line.end(), ',');
-        // If expected_num_commas was not given (i.e., is the default value of -1), it is equivalent to one less than
-        // the number of elements in parsed_line
-        if (expected_num_commas == -1)
-            expected_num_commas = line_size - 1;
-        // If expected_num_commas is given, we are only parsing some of the line, so the size of parsed_line is
-        // unrelated
+    if (separator == ',') {
+        int num_commas = std::count(line.begin(), line.end(), separator);
         if (expected_num_commas > num_commas) {
             std::string error = "Error: Expected " + std::to_string(line_size - 1) +
                                 " commas while reading file; but " + std::to_string(num_commas) + " were found";
@@ -155,7 +150,7 @@ void checkForHeaderValues(std::string header_line) {
     // Header values from file
     std::size_t header_size = 6;
     std::vector<std::string> header_values(header_size, "");
-    splitString(header_line, header_values);
+    splitString(header_line, header_values, 5);
 
     std::vector<std::vector<std::string>> expected_values = {{"x"}, {"y"}, {"z"}, {"tm"}, {"tl", "ts"}, {"r", "cr"}};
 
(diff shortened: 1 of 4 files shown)


src/CAparsefiles.cpp
```diff
@@ -361,28 +377,3 @@ std::string parseCoordinatePair(std::string line, int val) {
     SplitStr = std::regex_replace(SplitStr, r, "");
     return SplitStr;
 }
-
-// Separate ReadLine at commas and place components into ParsedLine, checking to make sure that the size of ParsedLine
-// is equal to the number of comma-separated values in ReadLine
-void parseCommaSeparatedArgs(std::string AnalysisFile, std::string ReadLine, std::vector<std::string> &ParsedLine) {
```
streeve, line 367: Remove the declaration too?

src/CAparsefiles.cpp
```diff
@@ -121,16 +121,32 @@ double getInputDouble(std::string val_input, int factor = 0) {
     return DoubleFromString;
 }
 
-// Given a line "s", parse at the commas and return the parsed values as strings in "ParsedLine"
-// If AllColumns = true, return all 6 values; otherwise, only parse the first 3 commas/values
-void splitString(std::string line, std::vector<std::string> &parsed_line, std::string separator = ",") {
+// Given a string ("line"), parse at "separator" (commas used by default)
+// Modifies "parsed_line" to hold the comma separated values
+void splitString(std::string line, std::vector<std::string> &parsed_line, std::string separator = ",",
```
streeve, line 126: Is it slower to return the vector instead of modify? Seems easier. If we keep it this way I don't think you need `expected_num_commas` - just use the size of the vector

streeve (changes requested):

src/CAparsefiles.cpp
```diff
@@ -121,16 +121,32 @@ double getInputDouble(std::string val_input, int factor = 0) {
...
+                 int expected_num_commas = -1) {
     std::size_t line_size = parsed_line.size();
+    // If reading a comma-separated line, make sure the right number of commas are present
+    if (separator == ",") {
```
streeve, line 130: I don't think this if block is necessary - an error for not finding enough values to fill the vector is reasonable, but `expected_num_commas` is never even used

MattRolchigo: `expected_num_commas` is used though. In most cases, the number of commas expected is the size of the line minus 1 (in this case, expected_num_commas is not used). But in one specific case, the number of commas expected is not equivalent to the size of the line - during the call on CAinitialize.cpp line 499, only the first 3 components of the line are to be read, despite 5 commas being expected on the line. I'll rewrite this a little so it's more clear

src/CAparsefiles.cpp
```diff
@@ -121,16 +121,27 @@ double getInputDouble(std::string val_input, int factor = 0) {
...
-// If AllColumns = true, return all 6 values; otherwise, only parse the first 3 commas/values
-void splitString(std::string line, std::vector<std::string> &parsed_line, std::string separator = ",") {
+// Given a string ("line"), parse at "separator" (commas used by default)
+// Modifies "parsed_line" to hold the comma separated values
+// expected_num_commas may be larger than the size of parsed_line, if only a portion of the line is being parsed
+void splitString(std::string line, std::vector<std::string> &parsed_line, int expected_num_commas,
+                 char separator = ',') {
     std::size_t line_size = parsed_line.size();
+    // If reading a comma-separated line, make sure the right number of commas are present
+    if (separator == ',') {
```
streeve, line 131: Should be for any separator

src/CAparsefiles.cpp
```diff
@@ -121,16 +121,25 @@ double getInputDouble(std::string val_input, int factor = 0) {
     return DoubleFromString;
 }
 
-// Given a line "s", parse at the commas and return the parsed values as strings in "ParsedLine"
-// If AllColumns = true, return all 6 values; otherwise, only parse the first 3 commas/values
-void splitString(std::string line, std::vector<std::string> &parsed_line, std::string separator = ",") {
+// Given a string ("line"), parse at "separator" (commas used by default)
+// Modifies "parsed_line" to hold the comma separated values
+// expected_num_commas may be larger than the size of parsed_line, if only a portion of the line is being parsed
+void splitString(std::string line, std::vector<std::string> &parsed_line, int expected_num_commas,
```
streeve, line 127: expected_num_values

streeve (approved):
Let's squash and merge

PR #94, opened 2022-08-25
Inline functions for octahedron calculations; CellCapture cleanup

### LINKED ISSUE #57: Create KOKKOS_INLINE_FUNCTION for calculation of critical diagonal lengths
Avoid redundant code in multiple subroutines by creating a separate inline function for critical diagonal length calculations

MattRolchigo:
Two inline functions (with associated unit tests) were created for decentered octahedron-related calculations and used in the appropriate locations in lieu of redundant code blocks. Variable declaration in CellCapture was moved to the first location where a variable is used. These changes resulted in GPU speedup and minor/negligible CPU speedup for the problems tested

Closes #57


src/CAupdate.hpp
```diff
@@ -12,6 +12,69 @@
...
+    DOCenter(3 * D3D1ConvPosition + 1) = GlobalY + 0.5;
+    DOCenter(3 * D3D1ConvPosition + 2) = GlobalZ + 0.5;
+}
+
+// For the newly active cell located at 1D array position D3D1ConvPosition (3D center coordinate of xp, yp, zp), update
+// CritDiagonalLength values for cell capture of neighboring cells. The octahedron has a center located at (cx, cy, cz)
+KOKKOS_INLINE_FUNCTION void calcCritDiagonalLength(int D3D1ConvPosition, float xp, float yp, float zp, float cx,
+                                                   float cy, float cz, NList NeighborX, NList NeighborY,
+                                                   NList NeighborZ, int MyOrientation, ViewF GrainUnitVector,
+                                                   ViewF CritDiagonalLength) {
```
streeve, line 29: Should pass Views by ref unless there's a clear reason not to. May not make a big difference, but better in general

unit_test/tstKokkosInit.hpp
```diff
@@ -421,9 +421,9 @@ void testCellTypeInit_NoRemelt() {
     // grid, these critical diagonal lengths were calculated to be the expected values required for
     // the octahedron to engulf the centers of each neighboring cells
     std::vector<float> CritDiagonalLength_Expected{
-        1.7009771, 2.1710062, 1.9409304, 1.9225504, 2.2444904, 2.6762383, 2.7775407, 2.6560771, 2.1579251,
```
streeve, line 424: Is this double/float differences? Seems concerning

streeve (commented):
Really nice improvement - couple small things

src/CAghostnodes.cpp
```diff
@@ -247,99 +247,12 @@ void GhostNodes2D(int, int, int NeighborRank_North, int NeighborRank_South, int
...
-                            double ParaT = (normx * x0 + normy * y0 + normz * z0) /
-                                           (normx * Diag1X + normy * Diag1Y + normz * Diag1Z);
-                            float CDLVal = pow(
-                                pow(ParaT * Diag1X, 2.0) + pow(ParaT * Diag1Y, 2.0) + pow(ParaT * Diag1Z, 2.0), 0.5);
-                            CritDiagonalLength((long int)(26) * CellLocation + (long int)(n)) = CDLVal;
-                        }
+                        calcCritDiagonalLength(CellLocation, xp, yp, zp, DOCenterX, DOCenterY, DOCenterZ, NeighborX,
+                                               NeighborY, NeighborZ, MyOrientation, GrainUnitVector,
+                                               CritDiagonalLength);
+                        // Change cell type to active only after CritDiagonalLength values were updated (for consistency
```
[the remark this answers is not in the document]
MattRolchigo: I mostly put that there as a rationale for why I moved the location of "CellType(GlobalCellLocation) = Active" from before the critical diagonal length calculations to after, since that's a change that was only tangentially related to adding the inline function

unit_test/tstKokkosInit.hpp
MattRolchigo: Yeah, the last few decimal points are different because of the previous switching back and forth between floats and doubles in the critical diagonal length calculations. For example, the first entry in that list of expected values went from 1.7009771 to 1.7009791. Didn't seem to affect the qualitative grain structure/texture results or parameters like volume fraction of nucleated grains for the example problems or for any of the larger problems I tested this with

src/CAghostnodes.cpp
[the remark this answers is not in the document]
streeve: Sure, it's just referencing a previous state of the code that didn't really affect anything. We've discussed it here, so I don't think it's necessary in line

streeve (approved):

PR #96, opened 2022-08-25
Allow printing of any RVE size without analysis script

MattRolchigo:
ExaCA previously only supported printing RVE data of a specific size without having to run the analysis script

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -34,7 +34,8 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
                        int &PrintDebug, bool &PrintMisorientation, bool &PrintFinalUndercoolingVals,
                        bool &PrintFullOutput, int &NSpotsX, int &NSpotsY, int &SpotOffset, int &SpotRadius,
                        bool &PrintTimeSeries, int &TimeSeriesInc, bool &PrintIdleTimeSeriesFrames,
-                       bool &PrintDefaultRVE, double &RNGSeed, bool &BaseplateThroughPowder, double &PowderDensity) {
+                       bool &PrintDefaultRVE, double &RNGSeed, bool &BaseplateThroughPowder, double &PowderDensity,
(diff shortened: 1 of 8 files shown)


src/CAprint.cpp
```diff
@@ -565,16 +565,13 @@ void PrintFinalUndercooling(std::string BaseFileName, std::string PathToOutput,
 // microstructure, and centered in the simulation domain in X and Y Default RVE size is 0.5 by 0.5 by 0.5 mm
 void PrintExaConstitDefaultRVE(std::string BaseFileName, std::string PathToOutput, int nx, int ny, int nz,
                                ViewI3D_H LayerID_WholeDomain, ViewI3D_H GrainID_WholeDomain, double deltax,
-                               int NumberOfLayers) {
-
-    // Determine the size, in CA cells of the RVE
-    long int RVESize = std::lrint(0.0005 / deltax);
```
streeve, line 571: Maybe rename `PrintRVE` `PrintExaConstitRVE` or since it doesn't assume a size

src/CAinitialize.cpp
```diff
@@ -439,6 +441,14 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
         PrintDefaultRVE = false;
         if (!(OptionalInputsRead_ProblemSpecific[2].empty()))
             PrintDefaultRVE = getInputBool(OptionalInputsRead_ProblemSpecific[2]);
+        // If printing RVE data, specify RVE size (in cells per side). Otherwise, RVE is 0.5 mm per side
+        if (!(OptionalInputsRead_ProblemSpecific[5].empty()))
```
streeve, line 445: Maybe add to some of the input files

src/CAprint.cpp
MattRolchigo: "Default" in this case still refers to the default location in the domain - but I'll clarify this in the comment above the subroutine definition

streeve (approved):

PR #98, opened 2022-08-30
Removed unused Melted variable and field

MattRolchigo:
Boolean array "Melted" wasn't used for anything, as whether or not a cell was part of the melt pool footprint was already indicated by LayerID (-1 if no, >= 0 if yes, with the exact value depending on the layer number)


src/CAprint.cpp
```diff
@@ -215,24 +214,15 @@ void PrintExaCAData(int id, int layernumber, int np, int nx, int ny, int nz, int
         ViewF3D_H UndercoolingCurrent_WholeDomain(
             Kokkos::ViewAllocateWithoutInitializing("UndercoolingCurrent_WholeDomain"), 0, 0, 0);
 
-        if ((PrintTimeSeries) || (PrintMisorientation) || (PrintDebug == 2) || (PrintFullOutput) || (PrintDefaultRVE)) {
```
streeve, line 218: Why is this no longer conditional?

MattRolchigo: Because LayerID is used in several locations in place of the removed Melted array, it is now needed for all printing possibilities. GrainID is collected in all except exactly one very specific (and nearly never used) combination of true/false/debug values for printing, so I decided that should just be collected in all cases as well to avoid confusing logic with 5 bool variable checks

streeve (approved):

PR #102, opened 2022-09-02
Updated tl/cr interpolation and added unit test

### LINKED ISSUE #95: Interpolation of temperature data fails in certain geometries for reduced/no remelt temp init

MattRolchigo:
Ported interpolation of tl and cr used when reading temperature data but not using the remelt logic to GPU, improved the selection of points associated with the interpolation, and added a unit test to test the updated interpolation.

ZMinLayer and ZMaxLayer needed to be converted into kokkos views for access on the device following the updated (device) interpolation: host and device copies are maintained as they are used on both. Unit tests that previously did not require kokkos but used ZMinLayer and ZMaxLayer were moved accordingly

Closes #95


streeve (commented):
This really could use a review from @colemanjs

MattRolchigo:
We should merge this soon before it falls too far behind the master branch...

streeve:
> We should merge this soon before it falls too far behind the master branch...

I can do another read through and I'm not blocking, but would prefer another review. You can always rebase on master after every few merges

PR #106, opened 2022-09-08
Separate ZBound_Low and ZBound_High functions for remelt/no remelt

MattRolchigo:
Removes subroutine DomainShiftAndResize_NoRemelt, replaces it with separate versions of the bounds calculations with and without remelting. LayerID, rather than CellType, is now used to determine a layer bounds for simulations without remelting - this yields the same result, but is done in preparation for CellType only being defined for a given layer, rather than for every cell in the multilayer domain

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -850,7 +850,7 @@ void ReadTemperatureData(int id, double &deltax, double HT_deltax, int &HTtoCAra
 int calcZBound_Low_Remelt(std::string SimulationType, int LayerHeight, int layernumber, float *ZMinLayer, float ZMin,
                           double deltax) {
 
-    int ZBound_Low;
+    int ZBound_Low = -1; // assign dummy initial value
     if (SimulationType == "S") {
         // lower bound of domain is an integer multiple of the layer spacing, since the temperature field is the
         // same for every layer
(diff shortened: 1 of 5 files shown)


src/CAinitialize.cpp
```diff
@@ -847,38 +847,51 @@ void ReadTemperatureData(int id, double &deltax, double HT_deltax, int &HTtoCAra
...
+        KOKKOS_LAMBDA(const int &D3D1ConvPosition, int &lmin) {
+            if (LayerID(D3D1ConvPosition) == layernumber) {
+                // Check Z position of this active cell
+                int RankZ = D3D1ConvPosition / (MyXSlices * MyYSlices);
+                if (RankZ < lmin)
+                    lmin = RankZ;
+            }
+        },
+        Kokkos::Min<int>(NewMin));
+    MPI_Allreduce(&NewMin, &ZBound_Low, 1, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
```
streeve, line 885: No reduction for remelt because it's per layer?

MattRolchigo: Yeah, we always know the Z bounds when we have remelting because we know the lower and upper Z coordinates, which are just converted into bounds. Without remelting, because the layers overlap, we use a reduction to figure out which cells are at the top and bottom of the layer and also don't remelt as part of another layer

streeve (approved):

PR #108, opened 2022-09-13
Added timer for steering vector creation

MattRolchigo:
The time spent in creation of the steering vector is now reported as a quantity separate from the time spent performing cell capture; previously, the time spent in cell capture was the sum of the steering vector creation + cell capture time


streeve (approved):
Looks good

PR #109, opened 2022-09-13
WIP: Resize sv

MattRolchigo:


streeve:
Should we close this?

MattRolchigo:
Yeah, sadly I don't think that attempting to avoid rebuilding the steering vector each time step is going to help us with performance

PR #110, opened 2022-09-20
Avoid parallel_reduce in JumpTimeStep if not needed

MattRolchigo:
Previously, calling the parallel_reduce with no cells to check would sometimes result in an error, as a large integer would be returned from the parallel_reduce. Now, the parallel_reduce is only called if a given MPI rank has cells to be checked


streeve (approved):
Looks good

PR #111, opened 2022-09-20
Updated example problems and README to include problems with remelting

MattRolchigo:


.github/workflows/CI.yml
```diff
@@ -87,8 +87,10 @@ jobs:
         run: |
           $HOME/exaca/bin/ExaCA-Kokkos Inp_SmallDirSolidification.txt
           $HOME/exaca/bin/ExaCA-Kokkos Inp_SmallSpotMelt.txt
+          $HOME/exaca/bin/ExaCA-Kokkos Inp_SmallSpotMelt_RM.txt
```
streeve, line 90: Maybe we only run examples with 2 ranks? We're up to about 5 min with debug on this PR and the unit test suite is starting to cover quite a bit

examples/Inp_SmallSpotMelt_RM.txt
```diff
@@ -0,0 +1,27 @@
+Test problem simulating solidification of a series of overlapping spots, with remelting of spots (Problem type 'SM')
+*****
+Problem type: SM
```
streeve, line 3: Worth considering if we should rename old files as "NoRemelt"

examples/README.md
```diff
@@ -1,5 +1,5 @@
 # ExaCA problem types and auxiliary files
-ExaCA currently can model three types of problems:
+ExaCA currently can model five types of problems:
```
streeve, line 2: I would still say 3, with and without remelting

streeve (changes requested):

.github/workflows/CI.yml
MattRolchigo: Fair enough, we could also turn debug printing/intermediate file output (which is a significant portion of the run time for these tiny examples) off for the 1 rank runs of SmallDirSolidification and SmallSpotMelt. I agree though, for the remelting spot test problem, no intermediate output + 2 ranks should be the only thing we do

examples/Inp_SmallSpotMelt_RM.txt
MattRolchigo: I'm reluctant to change the names of existing example problems - plus, since only 2 of the problems use remelting, I think it's reasonable to consider "no remelting" as the default for tests and only having "_RM" at the end of the examples that do consider remelting to avoid having unnecessarily long file names

.github/workflows/CI.yml
```diff
@@ -87,7 +87,6 @@ jobs:
         run: |
           $HOME/exaca/bin/ExaCA-Kokkos Inp_SmallDirSolidification.txt
```
streeve, line 88: Should consider removing the other 1 rank runs (separately if you prefer)

examples/README.md
```diff
@@ -8,7 +8,7 @@ ExaCA currently can model five types of problems:
...
+* Problem types SM and RM modify problem types S and R to include multiple melting and solidification events per cell. For problem types S and R all cells that will eventually undergo melting are initialized as liquid, and only the final time that a given cell goes below the liquidus temperature is considered. To obtain the most accurate results, all melting and solidification events should be considered; however, for some problem geometries, the microstructure resulting from only considering the final solidification event in each cell is a reasonable approximation (and faster))
```
streeve, line 11: ))

streeve (approved):

Issue #56, opened 2022-03-25
Consistent naming of cell coordinates


MattRolchigo:
Simplify and make consistent the labeling of 1D cell coordinates used for view access, as well as the corresponding 3D cell coordinates used to denote locations within each MPI rank's subdomain, the overall simulation domain, and each individual layer's portion of the simulation domain (for multilayer solidification problems)

PR #122, opened 2022-10-17
Removed 2D domain decomposition variables and functions

MattRolchigo:
No longer give the option of 2D domain decomposition schemes, simplifying the code and improving performance for simulations using 1D domain decomposition (as 2D decomposition was rarely shown to be faster than 1D, particularly on GPU backends).


src/CAupdate.cpp
```diff
@@ -317,7 +303,7 @@ void CellCapture(int, int np, int, int DecompositionStrategy, int, int, int MyXS
                             // Only proceed if CellType was previously liquid (this current thread changed the value to
                             // TemporaryUpdate)
                             if (OldCellTypeValue == Liquid) {
-                                int GlobalX = RankX + MyXOffset;
+                                int GlobalX = RankX;
```
streeve, line 306: Just use `RankX`?

streeve (commented):
First pass

src/CAupdate.cpp
MattRolchigo: Yeah that's probably better for now, though this also ties into #56

PR #123, opened 2022-10-18
Create interfacial response function class

streeve:
Needs to be tested on the device

### DIFF
--- a/src/CAinitialize.hpp
+++ b/src/CAinitialize.hpp
@@ -13,18 +13,18 @@
 #include <string>
 #include <vector>
 
-void InputReadFromFile(int id, std::string InputFile, std::string &SimulationType, int &DecompositionStrategy,
-                       double &AConst, double &BConst, double &CConst, double &DConst, double &FreezingRange,
-                       double &deltax, double &NMax, double &dTN, double &dTsigma, std::string &OutputFile,
-                       std::string &GrainOrientationFile, int &TempFilesInSeries, std::vector<std::string> &temp_paths,
-                       double &HT_deltax, bool &RemeltingYN, double &deltat, int &NumberOfLayers, int &LayerHeight,
-                       std::string &SubstrateFileName, float &SubstrateGrainSpacing, bool &UseSubstrateFile, double &G,
-                       double &R, int &nx, int &ny, int &nz, double &FractSurfaceSitesActive, std::string &PathToOutput,
-                       int &PrintDebug, bool &PrintMisorientation, bool &PrintFinalUndercoolingVals,
-                       bool &PrintFullOutput, int &NSpotsX, int &NSpotsY, int &SpotOffset, int &SpotRadius,
-                       bool &PrintTimeSeries, int &TimeSeriesInc, bool &PrintIdleTimeSeriesFrames,
-                       bool &PrintDefaultRVE, double &RNGSeed, bool &BaseplateThroughPowder, double &PowderDensity,
-                       int &RVESize);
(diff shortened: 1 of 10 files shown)


src/runCA.cpp
```diff
@@ -278,9 +279,7 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
                NumberOfSolidificationEvents, LayerTimeTempHistory);
 
     // Normalize solidification parameters
-    AConst = AConst * deltat / deltax;
-    BConst = BConst * deltat / deltax;
-    CConst = CConst * deltat / deltax;
+    irf.normalize(deltat, deltax);
```
MattRolchigo, line 282: This could actually be called inside of InputReadFromFile after these parameters have been initialized, since they aren't used or changed at any other point during initialization. Also I'm now realizing that DConst should also be normalized as well... since it is 0 for the material we've been using, it didn't matter, but that's a bug in ExaCA

streeve: Good point - maybe better to do that in a separate PR first to make it obvious?

src/CAinitialize.hpp
```diff
@@ -13,18 +13,18 @@
...
-                       std::string &GrainOrientationFile, int &TempFilesInSeries, std::vector<std::string> &temp_paths,
-                       double &HT_deltax, bool &RemeltingYN, double &deltat, int &NumberOfLayers, int &LayerHeight,
-                       std::string &SubstrateFileName, float &SubstrateGrainSpacing, bool &UseSubstrateFile, double &G,
-                       double &R, int &nx, int &ny, int &nz, double &FractSurfaceSitesActive, std::string &PathToOutput,
-                       int &PrintDebug, bool &PrintMisorientation, bool &PrintFinalUndercoolingVals,
-                       bool &PrintFullOutput, int &NSpotsX, int &NSpotsY, int &SpotOffset, int &SpotRadius,
-                       bool &PrintTimeSeries, int &TimeSeriesInc, bool &PrintIdleTimeSeriesFrames,
-                       bool &PrintDefaultRVE, double &RNGSeed, bool &BaseplateThroughPowder, double &PowderDensity,
-                       int &RVESize);
+std::string InputReadFromFile(int id, std::string InputFile, std::string &SimulationType, int &DecompositionStrategy,
```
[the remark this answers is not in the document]
MattRolchigo: Well now that I think about it, the material file is only defined locally inside of `InputReadFromFile`, while all of the other file names are initialized in `InputReadFromFile` but are passed into the subroutine by reference and used elsewhere. So for consistency, it's probably better than MaterialFile is another string passed by reference into `InputReadFromFile`, with the construction of the IRF class and associated initialization of FreezingRange, A, B, C, and D done inside of runCA.cpp where you have it now

src/runCA.cpp
streeve: Done in #125

src/CAinitialize.hpp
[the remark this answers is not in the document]
streeve: Agreed - done

src/runCA.cpp
streeve: Also improved in 2fc276b736fba61f9ec6f5b3f87f9ed58db6a164

MattRolchigo (approved):

MattRolchigo:
I'm ready to merge once that conflict is fixed

MattRolchigo (changes requested):
I'm ready to merge once that conflict is fixed

streeve:
So no requested changes?

MattRolchigo:
Nope, I'm good

MattRolchigo (approved):

PR #125, opened 2022-10-18
Normalize D constant for IRF 

streeve:
not an issue so far since D=0 for Inconel


streeve:
@MattRolchigo weird side effect here..

MattRolchigo:
Did this pass locally and fail on the CI only? I tried changing that line of the code myself and recompiling/testing on my computer and all the tests passed... I just reran the CI and it worked without an issue, so there's possibly some extremely rare race condition in the test itself

MattRolchigo (approved):

PR #126, opened 2022-10-19
Halo combined comm buffers and struct

MattRolchigo:
Add new data structures for halo exchange; add class to hold communication buffers

~I get an "undefined symbols" error during compilation, though I'm not sure why - maybe something in the cmake file?~

### DIFF
--- a/src/CAghostnodes.hpp
+++ b/src/CAghostnodes.hpp
@@ -6,38 +6,14 @@
 #ifndef EXACA_GHOST_HPP
 #define EXACA_GHOST_HPP
 
+#include "CAhalo.hpp"
 #include "CAtypes.hpp"
 
 #include <Kokkos_Core.hpp>
 
-// Load data (GrainID, DOCenter, DiagonalLength) into ghost nodes if the given RankY is associated with a 1D halo region
-KOKKOS_INLINE_FUNCTION void loadghostnodes(const double GhostGID, const double GhostDOCX, const double GhostDOCY,
-                                           const double GhostDOCZ, const double GhostDL, const int BufSizeX,
-                                           const int MyYSlices, const int RankX, const int RankY, const int RankZ,
-                                           const bool AtNorthBoundary, const bool AtSouthBoundary,
-                                           Buffer2D BufferSouthSend, Buffer2D BufferNorthSend) {
-
-    if ((RankY == 1) && (!(AtSouthBoundary))) {
-        int GNPosition = RankZ * BufSizeX + RankX;
-        BufferSouthSend(GNPosition, 0) = GhostGID;
-        BufferSouthSend(GNPosition, 1) = GhostDOCX;
-        BufferSouthSend(GNPosition, 2) = GhostDOCY;
-        BufferSouthSend(GNPosition, 3) = GhostDOCZ;
-        BufferSouthSend(GNPosition, 4) = GhostDL;
-    }
-    else if ((RankY == MyYSlices - 2) && (!(AtNorthBoundary))) {
-        int GNPosition = RankZ * BufSizeX + RankX;
-        BufferNorthSend(GNPosition, 0) = GhostGID;
-        BufferNorthSend(GNPosition, 1) = GhostDOCX;
-        BufferNorthSend(GNPosition, 2) = GhostDOCY;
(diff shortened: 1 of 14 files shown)


MattRolchigo:
The change set looks bigger than it is because it was based from #122 instead of master, most of the changes here are from that PR

streeve:
> I get an "undefined symbols" error during compilation

You were just missing the destructor

src/Halo.cpp
```diff
@@ -0,0 +1,47 @@
+// Copyright 2021 Lawrence Livermore National Security, LLC and other ExaCA Project Developers.
+// See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: MIT
+
+#include "Halo.hpp"
```
streeve, line 6: I think these files should just be name CAhalo.hpp and also include the free functions in ghostnodes

src/Halo.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <Kokkos_Core.hpp>
+
+//-constructor: buffers for ghost node data should be initialized with zeros
+Halo::Halo(const int BufSizeX, const int BufSizeZ) {
+
+    // Buffers for sending data in the +/- Y directions (used for 1D decomposition)
+    BufferSouthSend = ViewBuffer("BufferNS", 5 * BufSizeX * BufSizeZ);
```
streeve, line 13: Extra 5?

src/Halo.cpp
```diff
@@ -0,0 +1,47 @@
...
+    BufferSouthRecv = BufferSouthSend;
```
streeve, line 14: This is not going to go well.. they should all be created separately

src/CAtypes.hpp
```diff
@@ -37,8 +35,15 @@ typedef Kokkos::View<float ***, layout, Kokkos::HostSpace> ViewF3D_H;
 typedef Kokkos::View<int *, layout, Kokkos::HostSpace> ViewI_H;
 typedef Kokkos::View<int **, layout, Kokkos::HostSpace> ViewI2D_H;
 typedef Kokkos::View<int ***, layout, Kokkos::HostSpace> ViewI3D_H;
-typedef Kokkos::View<double **, layout, Kokkos::HostSpace> Buffer2D_H;
 
 typedef Kokkos::Array<int, 26> NList;
 
+struct Buffer {
```
streeve, line 41: Maybe `CommData` or something?

src/Halo.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <vector>
+
+// Class specifying ExaCA buffers for halo exchange operations
+
+class Halo {
+
+  public:
```
streeve, line 23: May as well make it a struct if everything is going to be public

src/CAghostnodes.hpp
```diff
@@ -7,125 +7,37 @@
...
 
 #include "CAtypes.hpp"
+#include "Halo.hpp"
 
 #include <Kokkos_Core.hpp>
 
 // Load data (GrainID, DOCenter, DiagonalLength) into ghost nodes if the given RankY is associated with a 1D halo region
-KOKKOS_INLINE_FUNCTION void loadghostnodes(const double GhostGID, const double GhostDOCX, const double GhostDOCY,
-                                           const double GhostDOCZ, const double GhostDL, const int BufSizeX,
+KOKKOS_INLINE_FUNCTION void loadghostnodes(const int GhostGID, const float GhostDOCX, const float GhostDOCY,
```
streeve, line 15: I would consider making all of these member functions of `Halo`

streeve (changes requested):
I still think separating the new comm buffer type and the new halo class is worth the time, but this is close to ready probably

MattRolchigo:
Current status: 

- Compiles on Serial and Pthreads backends
- Compiles with warnings with Cuda backend ("calling a __host__ function("Halo::~Halo") from a __host__ __device__ function("") is not allowed", and ("Implicit capture of 'this' in extended lambda expression")
- Compiles with warnings with HIP backend ("capture host side class data member by this pointer in device or host device lambda function may result in invalid memory access if this pointer is not accessible on device side [-Wgpu-maybe-wrong-side]") - BufferSouthSend in CAhalo.cpp

PR #134, opened 2022-10-25
New function for calculating grain misorientation

MattRolchigo:
Replace duplicated code from CA and grain analysis print routines with a new function


streeve (approved):
Any simple test for this?

MattRolchigo:
I did run the small directional solidification example with the master branch and this one, and the misorientation .vtk plot was identical as was the mean misorientation when running the analysis executable

PR #135, opened 2022-10-25
New function for identifying unique GrainIDs

MattRolchigo:
Removes duplicated code from grain analysis print routine


analysis/src/GAutils.cpp
```diff
@@ -527,3 +527,14 @@ void ParseAnalysisFile(std::string AnalysisFile, std::string RotationFilename, i
...
     Analysis.close();
 }
+
+// Given an input vector of integer Grain ID values, return an output vector consisting of the unique Grain ID values,
+// sorted from lowest to highest
+std::vector<int> FindUniqueGrains(const std::vector<int> GrainIDVector) {
+    std::vector<int> UniqueGrainIDVector = GrainIDVector;
+    std::sort(UniqueGrainIDVector.begin(), UniqueGrainIDVector.end());
+    std::vector<int>::iterator it;
+    it = std::unique(UniqueGrainIDVector.begin(), UniqueGrainIDVector.end());
```
streeve, line 537: If we require newer versions, we can start to transition this kind of thing to the device: https://kokkos.github.io/kokkos-core-wiki/API/algorithms/std-algorithms/all/StdUnique.html

streeve (approved):

PR #136, opened 2022-10-27
New analysis executable for AMBench microstructure

MattRolchigo:


analysis/src/CMakeLists.txt
```diff
@@ -8,7 +8,20 @@ set(EXAGA_SOURCES
...
+set(EXAGA_AMB_SOURCES
+    runGA_AMB.cpp
+    GAprint_AMB.cpp
+    GAutils.cpp
+)
 
 add_executable(grain_analysis ${EXAGA_SOURCES})
 target_link_libraries(grain_analysis ExaCA)
 install(TARGETS grain_analysis DESTINATION ${CMAKE_INSTALL_BINDIR})
+add_executable(amb_grain_analysis ${EXAGA_AMB_SOURCES})
```
streeve, line 25: I would go with `grain_analysis_amb`

analysis/src/GAutils.hpp
```diff
@@ -21,7 +21,9 @@ std::string parseCoordinatePair(std::string line, int val);
 int FindTopOrBottom(int ***LayerID, int XLow, int XHigh, int YLow, int YHigh, int nz, int L, std::string HighLow);
 
 // These are used in reading/parsing ExaCA microstructure data
-void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax, int &NumberOfLayers);
+static std::vector<double> DEFAULT_VECTOR;
```
streeve, line 24: What

analysis/src/runGA_AMB.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+// The name of the ExaCA output analysis file (.txt, located in ExaCA/analysis), the ExaCA microstructure file (.vtk),
+// and the ExaCA log file (.log) is given on the command line - these files should all have the same name, other than
+// the file extension
+int main(int argc, char *argv[]) {
```
streeve, line 24: Is there a quick run you can add to the CI?

streeve (changes requested):
Not sure if the print functions need to be separate files - there's already an ExaConstit-specific function in the "main" analysis

analysis/src/GAutils.hpp
streeve: Seems much simpler to return the vector and ignore if not needed

analysis/src/GAutils.cpp
```diff
@@ -103,6 +118,15 @@ void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax
     }
     std::cout << "Number of layers in microstructure data: " << NumberOfLayers << std::endl;
     InputDataStream.close();
+
+    // If log file contained the X, Y, Z bounds of the domain and they're used in the analysis, convert them
```
streeve, line 122: Won't the parsing above fail if the bounds aren't included?

MattRolchigo: Yes - only ExaCA simulation results starting with this commit are compatible with this new analysis executable. I don't really see a reason why I'd need to run this new analysis script using any of the older datasets, though I should probably throw a specific error if I accidentally do try to run this with an old dataset

MattRolchigo:
Regarding the print functions, since the old analysis script exclusively uses print functions from GAprint, and the new analysis script exclusively uses print functions from GAprint_AMB, it seemed like having separate files was a more logical way to divide things. There is probably a better way to deal with the redundancy between PrintExaConstitRVEData in GAprint and PrintExaConstitDefaultRVE in CAprint, but the new analysis script does not print ExaConstit data (for the ExaAM challenge problem, we're relying on PrintExaConstitDefaultRVE being called as part of running ExaCA)

MattRolchigo:
the addition of AnalyzeThickLeg.txt in analysis/examples was a mistake, that file has been removed. The new analysis executable does not require an input file, just a command line path/name of the dataset

analysis/src/GAprint_AMB.cpp
```diff
@@ -0,0 +1,359 @@
+// Copyright 2021 Lawrence Livermore National Security, LLC and other ExaCA Project Developers.
+// See the top-level LICENSE file for details.
+//
+// SPDX-License-Identifier: MIT
+#include "GAprint_AMB.hpp"
+#include "CAfunctions.hpp"
+#include "GAutils.hpp"
+
+#include <algorithm>
```
streeve, line 9: Do you really need all these headers? They're also duplicated from the header

analysis/src/GAutils.cpp
```diff
@@ -194,6 +234,24 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string
     }
 }
 
+// Ensure that the appropriate files exist
+void ParseFilenames_AMB(std::string BaseFileName, std::string &LogFile, std::string &MicrostructureFile,
```
streeve, line 238: Can't this call `ParseFilenames` and live in the AMB header?

analysis/src/GAprint_AMB.cpp
```diff
@@ -0,0 +1,359 @@
...
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+// Helper function for PrintDefaultCrossSections_AMB and PrintCrossSectionSeries_AMB
+void WriteEulerAngleData_AMB(std::ofstream &Grainplot, ViewI3D_H &GrainID, int i, int j, int k,
```
streeve, line 20: I feel like all of these functions should call a corresponding method from `GAprint` somewhere

analysis/src/GAprint_AMB.cpp
MattRolchigo: Probably not. They're in GAprint.cpp too and don't need to be, I'll remove them

analysis/src/GAutils.cpp
MattRolchigo: Despite the similar names, ParseFilenames and ParseFilenames_AMB are actually pretty different. ParseFilenames actually reads and parses an input file, while ParseFilenames_AMB just takes the command line input and checks that the files exist/are installed. Maybe renaming ParseFilenames_AMB to CheckInputFiles would be more appropriate to reflect this difference

analysis/src/GAprint_AMB.cpp
MattRolchigo: I'm not 100% sure what you mean, there's some repeated code that could be removed by rewriting/combining parts of WriteEulerAngleData_AMB and PrintCrossSectionOrientationData. But PrintXYCrossSectionSeries_AMB and PrintDefaultCrossSectionData_AMB are pretty distinct from anything done inside of GAprint, I don't see how that would work

analysis/src/GAutils.cpp
streeve: Rename sounds good to me

analysis/src/GAprint_AMB.cpp
streeve: I would refactor Euler then and rename the others to distinguish if they're that different

streeve (changes requested):

analysis/src/GAprint.cpp
```diff
@@ -326,46 +316,279 @@ void PrintPoleFigureData(bool *AnalysisTypes, std::string BaseFileName, int Numb
...
+    for (int n = 0; n < NumberOfGrains; n++) {
+        if (GrainAreas[n] >= MinGrainSize) {
+            AreaGrainsAboveThreshold += GrainAreas[n];
+            NumGrainsAboveThreshold++;
+        }
+    }
+
+    // What fraction of the cross-sectional area consists of grains large enough to be counted?
+    double FractAreaAboveThreshold =
+        static_cast<double>(AreaGrainsAboveThreshold) / static_cast<double>(CrossSectionSize);
```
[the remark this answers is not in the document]
MattRolchigo: Like a function that takes two ints as input parameters, casts them to doubles, and returns the quotient as a double? That seemed pretty specific to me at first, but I realized I actually do this operation multiple times in the code...

analysis/src/GAutils.cpp
```diff
@@ -192,8 +238,42 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string
...
+    }
+    else
+        RGBFilename = FilesRead[5];
+    // Check that file exists and was successfully installed
+    RGBFilename = checkFileInstalled(RGBFilename, 0);
 }
 
+// Ensure that the appropriate files exist - orientation files should be installed, other files should start with the
+// BaseFileName value given from the command line
+void CheckInputFiles_AMB(std::string BaseFileName, std::string &LogFile, std::string &MicrostructureFile,
```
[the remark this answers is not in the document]
streeve: The clutter seems worth not having this in GAutils to me

analysis/src/runGA_AMB.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    else {
+        BaseFileName = argv[1];
+        CheckInputFiles(BaseFileName, LogFile, MicrostructureFile, RotationFilename, EulerAnglesFilename, RGBFilename);
+    }
+    std::cout << "Performing analysis of " << MicrostructureFile << " , using the log file " << LogFile << std::endl;
+
+    // Initialize Kokkos
+    Kokkos::initialize();
+    {
```
streeve, line 41: You don't need this extra scope if you're using init/finalize

analysis/src/GAprint.cpp
```diff
@@ -326,46 +317,280 @@ void PrintPoleFigureData(bool *AnalysisTypes, std::string BaseFileName, int Numb
...
+    MisorientationLargeZ.open(MisorientationFileLargeZ);
+    GreenBlueFreq.open(GreenBlueFreqFile);
+
+    ViewF_H GrainMisorientationX = MisorientationCalc(NumberOfOrientations, GrainUnitVector, 0);
+    ViewF_H GrainMisorientationY = MisorientationCalc(NumberOfOrientations, GrainUnitVector, 1);
+    ViewF_H GrainMisorientationZ = MisorientationCalc(NumberOfOrientations, GrainUnitVector, 2);
+    float GrainMisorientation_small_sumx = 0, GrainMisorientation_small_sumy = 0, GrainMisorientation_small_sumz = 0;
+    float GrainMisorientation_large_sumx = 0, GrainMisorientation_large_sumy = 0, GrainMisorientation_large_sumz = 0;
+    float GreenBlueRatioSum = 0.0;
+    for (int n = 0; n < NumberOfGrains; n++) {
```
streeve, line 462: Is this ever enough of a bottleneck to do it on the device?

analysis/src/GAutils.cpp
```diff
@@ -538,3 +645,9 @@ std::vector<int> FindUniqueGrains(const std::vector<int> GrainIDVector) {
     UniqueGrainIDVector.resize(std::distance(UniqueGrainIDVector.begin(), it));
     return UniqueGrainIDVector;
 }
+
+// Cast two integers to double, divide them, and return the result
+double DivideInts(int Int1, int Int2) {
+    double quotient = static_cast<double>(Int1) / static_cast<double>(Int2);
+    return quotient;
```
streeve, line 652:
```suggestion
template <typename ReturnType, typename FirstType, typename SecondType>
ReturnType DivideInts(FirstType Int1, SecondType Int2) {
    return static_cast<ReturnType>(Int1) / static_cast<ReturnType>(Int2);
```

streeve: Then you can use this for floats/different input types too

streeve (commented):
Unless it's worth doing the size calculation on device, then printing separately, I think this is basically ready

analysis/src/GAprint.cpp
MattRolchigo: Since we just call it on individual cross-sections, there are never enough grains that it becomes a bottleneck. Some of the older analysis routines that we call on a volume of material could benefit from being done on the device, though

analysis/src/runGA_AMB.cpp
MattRolchigo: We do the same thing in runGA.cpp, I didn't want to change things up here

analysis/src/GAutils.cpp
MattRolchigo: I tried something this and couldn't get it to work, with compilation errors about `candidate template ignored: couldn't infer template argument 'ReturnType'`. I'm assuming I'd need a copy of the line `template <typename ReturnType, typename FirstType, typename SecondType>` in the .hpp file too, but even with that I get the error

analysis/src/GAutils.hpp
```diff
@@ -21,20 +21,30 @@ std::string parseCoordinatePair(std::string line, int val);
...
-                       bool NewOrientationFormatYN);
+                       std::vector<bool> &BimodalAnalysis, bool NewOrientationFormatYN,
+                       std::vector<std::string> &CSLabels);
 std::vector<int> FindUniqueGrains(const std::vector<int> GrainIDVector);
 
+void CheckInputFiles(std::string BaseFileName, std::string &LogFile, std::string &MicrostructureFile,
+                     std::string &RotationFilename, std::string &RGBFilename, std::string &EulerAnglesFilename);
+std::vector<int> FindUniqueGrains(std::vector<int> GrainIDVector);
+template <typename ReturnType, typename FirstType, typename SecondType>
+ReturnType DivideCast(FirstType Int1, SecondType Int2) {
```
streeve, line 46: On second glance, we should probably name them `Input1` and `Input2`

analysis/src/runGA_AMB.cpp
streeve: I'd rather fix it here and fix others later

streeve (commented):
I'll approve after current comments are resolved

streeve (approved):

PR #138, opened 2022-11-03
Writing and reading of ExaCA output data as binary

MattRolchigo:
Added the option to print ExaCA vtk data as binary data (instead of ASCII, which is currently used by default) to reduce the output file sizes. To support this, new functions were added to read and write binary data, and convert between little and big endian format, and a new unit test was written to test the binary input/output.

LayerID and GrainMisorientation data is now printed as "short int" data instead of "int" to reduce file size, this won't matter for GrainMisorientation which will never reach the lower/upper limits of this data type, and would only matter for LayerID in the unlikely case that more than 32,767 layers of microstructure were simulated

### DIFF
--- a/src/CAparsefiles.hpp
+++ b/src/CAparsefiles.hpp
@@ -8,6 +8,9 @@
 
 #include "CAtypes.hpp"
 
+#include <fstream>
+#include <iostream>
+#include <sstream>
 #include <string>
 
 void skipLines(std::ifstream &stream, std::string seperator);
@@ -36,5 +39,33 @@ void checkFileNotEmpty(std::string testfilename);
 void parseMaterialFile(std::string MaterialFile, double &AConst, double &BConst, double &CConst, double &DConst,
                        double &FreezingRange);
 std::string parseCoordinatePair(std::string line, int val);
+// Swaps bits for a variable of type SwapType
+template <typename SwapType>
+void SwapEndian(SwapType &var) {
+    // Cast var into a char array (bit values)
+    char *varArray = reinterpret_cast<char *>(&var);
+    // Size of char array
+    int varSize = sizeof(var);
+    // Swap the "ith" bit with the bit "i" from the end of the array
+    for (long i = 0; i < static_cast<long>(varSize / 2); i++)
+        std::swap(varArray[varSize - 1 - i], varArray[i]);
+}
+// Reads binary data of the type ReadType, optionally swapping the endian format
+template <typename ReadType>
+ReadType ReadBinaryData(std::ifstream &instream, bool SwapEndianYN = false) {
+    unsigned char temp[sizeof(ReadType)];
+    instream.read(reinterpret_cast<char *>(temp), sizeof(ReadType));
+    if (SwapEndianYN)
+        SwapEndian(temp);
+    ReadType readValue = reinterpret_cast<ReadType &>(temp);
+    return readValue;
+}
+// Parse space-separated ASCII data loaded into the string stream
+template <typename ReadType>
+ReadType ParseASCIIData(std::istringstream &ss) {
+    ReadType readValue;
+    ss >> readValue;
+    return readValue;
+}
 
 #endif
--- a/src/CAprint.cpp
+++ b/src/CAprint.cpp
@@ -5,13 +5,30 @@
 
 #include "CAprint.hpp"
 #include "CAfunctions.hpp"
+#include "CAparsefiles.hpp"
 
 #include "mpi.h"
 
-#include <cmath>
-#include <fstream>
+// Print Paraview file header data, for either binary or ASCII output
+void WriteHeader(std::ofstream &ParaviewOutputStream, std::string FName, bool PrintBinary, int nx, int ny, int nz,
+                 double deltax, double XMin, double YMin, double ZMin) {
 
-//*****************************************************************************/
+    if (PrintBinary)
+        ParaviewOutputStream.open(FName, std::ios::out | std::ios::binary);
+    else
+        ParaviewOutputStream.open(FName);
+    ParaviewOutputStream << "# vtk DataFile Version 3.0" << std::endl;
+    ParaviewOutputStream << "vtk output" << std::endl;
+    if (PrintBinary)
+        ParaviewOutputStream << "BINARY" << std::endl;
+    else
+        ParaviewOutputStream << "ASCII" << std::endl;
+    ParaviewOutputStream << "DATASET STRUCTURED_POINTS" << std::endl;
+    ParaviewOutputStream << "DIMENSIONS " << nx << " " << ny << " " << nz << std::endl;
+    ParaviewOutputStream << "ORIGIN " << XMin << " " << YMin << " " << ZMin << std::endl;
+    ParaviewOutputStream << "SPACING " << deltax << " " << deltax << " " << deltax << std::endl;
+    ParaviewOutputStream << std::fixed << "POINT_DATA " << nx * ny * nz << std::endl;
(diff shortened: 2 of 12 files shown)


src/CAparsefiles.hpp
```diff
@@ -35,5 +38,22 @@ void checkFileNotEmpty(std::string testfilename);
 void parseMaterialFile(std::string MaterialFile, double &AConst, double &BConst, double &CConst, double &DConst,
                        double &FreezingRange);
 std::string parseCoordinatePair(std::string line, int val);
+template <typename SwapType>
```
streeve, line 41: I think this needs a comment desciption

src/CAprint.cpp
```diff
@@ -345,10 +355,14 @@ void PrintCAFields(int nx, int ny, int nz, ViewI3D_H GrainID_WholeDomain, ViewI3
...
                 for (int i = 0; i < nx; i++) {
-                    Grainplot << GrainID_WholeDomain(k, i, j) << " ";
+                    if (PrintBinary)
+                        WriteBinaryData(Grainplot, GrainID_WholeDomain(k, i, j), true);
+                    else
+                        Grainplot << GrainID_WholeDomain(k, i, j) << " ";
                 }
             }
-            Grainplot << std::endl;
+            if (!(PrintBinary))
```
streeve, line 364: Is it bad to end on a newline for binary?

src/CAprint.cpp
```diff
@@ -318,25 +317,36 @@ void PrintCAFields(int nx, int ny, int nz, ViewI3D_H GrainID_WholeDomain, ViewI3
...
     std::ofstream Grainplot;
-    Grainplot.open(FName);
+    if (PrintBinary)
+        Grainplot.open(FName, std::ios::out | std::ios::binary);
+    else
+        Grainplot.open(FName);
     Grainplot << "# vtk DataFile Version 3.0" << std::endl;
     Grainplot << "vtk output" << std::endl;
-    Grainplot << "ASCII" << std::endl;
+    if (PrintBinary)
```
streeve, line 326: I would prefer to pass the `bool` & string, then move this `if` into the `writeData` function

src/CAprint.hpp
```diff
@@ -44,19 +49,29 @@ void PrintCAFields(int nx, int ny, int nz, ViewI3D_H GrainID_WholeDomain, ViewI3
...
                                int NumberOfLayers, int RVESize);
 void PrintIntermediateExaCAState(int IntermediateFileCounter, int layernumber, std::string BaseFileName,
                                  std::string PathToOutput, int ZBound_Low, int nzActive, int nx, int ny,
                                  ViewI3D_H GrainID_WholeDomain, ViewI3D_H CellType_WholeDomain, ViewF_H GrainUnitVector,
-                                 int NGrainOrientations, double deltax, float XMin, float YMin, float ZMin);
+                                 int NGrainOrientations, double deltax, float XMin, float YMin, float ZMin,
+                                 bool PrintBinary);
+// Write data as binary, with option to convert between big and small endian
+template <typename PrintType>
+void WriteBinaryData(std::ofstream &outstream, PrintType PrintValue, bool SwapEndianYN) {
```
streeve, line 70: Seems like the endian bool should have a default

streeve (changes requested):
Can you update the description? I know you added changes specifically for reduced precision in some places in addition to the binary option

src/CAprint.cpp
MattRolchigo: The newline was only necessary for the ASCII format (I believe, it might not have been necessary to begin with), it shouldn't be needed for the binary or ASCII really

src/CAparsefiles.hpp
streeve: I think this needs line by line comment descriptions

src/CAprint.cpp
```diff
@@ -318,25 +337,22 @@ void PrintCAFields(int nx, int ny, int nz, ViewI3D_H GrainID_WholeDomain, ViewI3
...
     // Print Layer ID data
-    Grainplot << "SCALARS LayerID int 1" << std::endl;
+    Grainplot << "SCALARS LayerID short 1" << std::endl;
     Grainplot << "LOOKUP_TABLE default" << std::endl;
     for (int k = 0; k < nz; k++) {
         for (int j = 0; j < ny; j++) {
             for (int i = 0; i < nx; i++) {
-                Grainplot << LayerID_WholeDomain(k, i, j) << " ";
+                short ShortLayerID = static_cast<short>(LayerID_WholeDomain(k, i, j));
+                if (PrintBinary)
```
streeve, line 348: Let's also move this `if` block into a general `writeData` function that does either binary or plain text

src/CAprint.cpp
streeve: Why not print for either or both then?

MattRolchigo: Turns out, the newline isn't actually needed for the ASCII format (that's something in the code that hasn't been touched in 2 years, and was probably in there "just in case")

streeve: Looks like the newline is necessary in some cases based on the CI failure

unit_test/tstInit.cpp
```diff
@@ -193,9 +201,47 @@ void testInputReadFromFile() {
...
+    float FloatData[5] = {-1.0, 0.0, 1.0, 2.0, 3.0};
+
+    // Write data as binary to be used as input
+    std::ofstream TestIntData;
+    TestIntData.open("TestIntData.txt", std::ios::out | std::ios::binary);
+    std::ofstream TestFloatData;
+    TestFloatData.open("TestFloatData.txt", std::ios::out | std::ios::binary);
+    for (int n = 0; n < 5; n++) {
+        // Write to files
+        WriteData(TestIntData, IntData[n], true, true);
```
streeve, line 225: Let's call this function with both binary and text (this test should catch the current CI error in addition to the analysis script)

src/CAprint.cpp
```diff
@@ -5,13 +5,30 @@
...
+#include "CAparsefiles.hpp"
 
 #include "mpi.h"
 
-#include <cmath>
-#include <fstream>
-
-//*****************************************************************************/
+// Print Paraview file header data, for either binary or ASCII output
+void OpenPrintHeader(std::ofstream &ParaviewOutputStream, std::string FName, bool PrintBinary, int nx, int ny, int nz,
```
streeve, line 13: `writeHeader`?

unit_test/tstInit.cpp
MattRolchigo: Yeah this error is from the fact that I didn't account for the removal of the newlines in the ASCII file read in `ReadASCIIField`, I'll have to fix that and add a test

MattRolchigo: `ReadASCIIField` should also account for the fact that vtk files printed prior to this commit will have the newlines, while the ones from this point on will not

src/CAprint.cpp
MattRolchigo: Yeah I forgot about the fact that there's multiple fields of vtk data in some files that need to be separated from others with header lines on new lines

MattRolchigo: So it turns out, printing the newline at the end of the field data messes up the binary read. `outstream.write((char *)&somebinaryvalue, somebinarysize)` followed by `outstream << "Next field header data << std::endl;` works, but having any other lines in there does not. But the `std::endl;` is still needed at the end of ASCII field data, since otherwise it don't print the next field's header data to a new line

streeve (approved):

PR #140, opened 2022-11-15
Add JSON for material input files

streeve:
Auto-download and include json (FetchContent requires new min CMake 3.11). Currently set up to enable turning this off entirely


MattRolchigo:
@streeve The compilation of this seems to fail based on the order in which the files are compiled... sometimes it works, sometimes I get the error that shows up in the CI. Right now "CAinterfacialresponse.hpp" is only file where nlohmann/json.hpp is included

streeve:
> The compilation of this seems to fail based on the order in which the files are compiled... sometimes it works, sometimes I get the error that shows up in the CI. Right now "CAinterfacialresponse.hpp" is only file where nlohmann/json.hpp is included

Should be fixed by #141

examples/Materials/Inconel625_Approx
```diff
@@ -0,0 +1,9 @@
+{
+   "function": "quadratic",
```
streeve, line 2: Rename `Inconel625_quadratic`?

streeve (commented):
I rewrote the IRF to avoid that if statement in the compute function (and to avoid scary enums/integers). A lot of it's annoying boilerplate, but I don't know a better way really. Looks like a lot of changes mostly because anything using the IRF is now templated and moved to the header

MattRolchigo (approved):

PR #142, opened 2022-11-18
Add core library header file

streeve:
Use in grain analysis

### DIFF
--- a/analysis/src/GAprint.cpp
+++ b/analysis/src/GAprint.cpp
@@ -2,8 +2,8 @@
 // See the top-level LICENSE file for details.
 //
 // SPDX-License-Identifier: MIT
+
 #include "GAprint.hpp"
-#include "CAfunctions.hpp"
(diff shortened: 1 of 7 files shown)


MattRolchigo:
So this is just to simplify the include statements, so only one header file for ExaCA subroutines needs to be included in each of the analysis/src/*.hpp files?

streeve:
> So this is just to simplify the include statements, so only one header file for ExaCA subroutines needs to be included in each of the analysis/src/*.hpp files?

Yeah, this will simplify another app using ExaCA as a library and the analysis is just our only example so far

MattRolchigo:
Sounds good to me

streeve:
> Sounds good to me

Cool, but do you "approve" of the changes?

MattRolchigo:
I think I have to be requested as a reviewer first

MattRolchigo (approved):

PR #143, opened 2022-11-19
Consolidation of layer bound check functions to simplify code

MattRolchigo:
Removal of calcZBound_Low_NoRemelt and calcZBound_Low_NoRemelt functions - code without remelting now uses the same bounds calculation functions as the remelting code. This will lead to iteration over more cells than needed for certain situations in simulations without remelting, but the effect on performance appears to be minor to negligible (<3% of runtime) and the effect on the result is minor (iterating over the extra cells per layer slightly biases the simulation towards growth of existing grains over nucleation/growth of new ones)


streeve (approved):
Makes sense

PR #145, opened 2022-11-30
Add JSON parsing for materials inputs

streeve:
- Adds optional JSON (retains previous plain text) for material IRF inputs
- New options for different interfacial response functions ~(factory method to generate at runtime)~


CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/ExaCAconfig.cmake"
   DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/ExaCA)
 
 # Install data files
-file(GLOB MATERIALS examples/Materials/Inconel625)
+file(GLOB MATERIALS examples/Materials/*)
```
MattRolchigo, line 22: Is it necessary that the material file used for a given simulation is installed? Or can it still be from a user-specified path in the input file? If a user had 50 material files that they generated themselves in addition to the 3 in the repo, this would install all 53; is installing that many files bad?

unit_test/tstInit.cpp
```diff
@@ -124,7 +125,8 @@ void testInputReadFromFile() {
                           PrintFinalUndercoolingVals, PrintFullOutput, NSpotsX, NSpotsY, SpotOffset, SpotRadius,
                           PrintTimeSeries, TimeSeriesInc, PrintIdleTimeSeriesFrames, PrintDefaultRVE, RNGSeed,
                           BaseplateThroughPowder, PowderDensity, RVESize, LayerwiseTempInit, PrintBinary);
-        InterfacialResponseFunction irf(MaterialFileName, deltat, deltax);
+#ifdef ExaCA_ENABLE_JSON
+        auto irf = IRFfactory::create(0, MaterialFileName, deltat, deltax);
```
streeve, line 129: Out of date

src/CAinterfacialresponse.hpp
```diff
@@ -82,6 +126,24 @@ struct InterfacialResponseFunction {
         D *= deltat / deltax;
     }
 
+    // Compute velocity from local undercooling.
+    KOKKOS_INLINE_FUNCTION
+    double compute(const double LocU) const {
+        double V;
+        switch (function) {
```
streeve, line 133: Why use switch when you want to break every time?

MattRolchigo: I thought that's how they're done, like this: https://www.w3schools.com/cpp/cpp_switch.asp

streeve: I really mean why not use `if`

MattRolchigo: It seemed like the case statement would be more efficient for compiler optimizations, though I didn't see a straight answer online about that for C++

src/CAinterfacialresponse.hpp
```diff
@@ -82,6 +126,19 @@ struct InterfacialResponseFunction {
...
 
+    // Compute velocity from local undercooling.
+    KOKKOS_INLINE_FUNCTION
+    double compute(const double LocU) const {
+        double V;
+        if (function == 0)
+            V = A * pow(LocU, 3.0) + B * pow(LocU, 2.0) + C * LocU + D;
+        else if (function == 1)
+            V = A * pow(LocU, 2.0) + B * LocU + C;
+        else
```
streeve, line 137: Would prefer `else if` here and an error in the `else`

MattRolchigo: Since this is inside of a kokkos inline function, we aren't able to use `throw std::runtime_error` in an `else` statement guarding against function values other than 0, 1, or 2

CMakeLists.txt
streeve: I think we agreed to leave this behavior for now and update separately if needed

streeve:
@MattRolchigo updated this a couple more times: 1) rebased to clean up the commits 2) made it possible to compile in an existing copy of the json library or auto-download if not found. Need your approval, then I'll merge

MattRolchigo (approved):

PR #149, opened 2022-12-07
Remove deprecated functions for temperature input

streeve:
available in 1.1

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -85,6 +85,10 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
     std::vector<std::string> RequiredInputs_ProblemSpecific, OptionalInputs_ProblemSpecific,
         DeprecatedInputs_ProblemSpecific;
     std::vector<bool> DeprecatedInputs_RequiredYN;
+    // Currently no deprecated inputs. Keeping variables in anticipation of more in the future.
+    DeprecatedInputs_ProblemSpecific.resize(0);
+    // If using deprecated inputs, which ones are required/optional
+    DeprecatedInputs_RequiredYN.resize(0);
     if (SimulationType == "RM") {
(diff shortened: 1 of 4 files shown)


MattRolchigo:
Should we create another release so that the error message can say something like "Missing temperatures input file, the old format was removed starting with version 1.2"?

src/CAinitialize.cpp
```diff
@@ -226,21 +226,9 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
     }
     for (int i = 0; i < NumRequiredInputs_ProblemSpecific; i++) {
         if (RequiredInputsRead_ProblemSpecific[i].empty()) {
-            if ((SimulationType == "R") && (i == 1)) {
-                if (id == 0)
-                    std::cout << "Missing temperature field assembly instructions : checking for deprecated "
-                                 "temperature inputs"
-                              << std::endl;
-                parseTemperatureInput_Old(DeprecatedInputs_ProblemSpecific, DeprecatedInputsRead_ProblemSpecific,
```
MattRolchigo, line 234: DeprecatedInputs_ProblemSpecific[0] through DeprecatedInputs_ProblemSpecific[5] should be removed too, since they're only used in the case of parsing the old temperature input. Also, DeprecatedInputs_ProblemSpecific[6] has been in there long enough that we can probably just remove it in this  PR or a related one as well (we have a comment that says "Fixme: Remove this optional input eventually" and a note about the Extra walls input no longer being used)

streeve:
> Should we create another release so that the error message can say something like "Missing temperatures input file, the old format was removed starting with version 1.2"?

I don't think that's enough to justify a release. The error will still say that the temperature input file is missing, right?

src/CAinitialize.cpp
streeve: Good point - done

MattRolchigo (approved):

PR #150, opened 2022-12-08
Consistent use of double for temperature data

MattRolchigo:
Code previously used `float` (or inconsistently used `float` and `double`) instead of `double` for intermediate quantities derived from the double precision temperature input data


src/CAinitialize.cpp
```diff
@@ -1575,9 +1575,9 @@ void TempInit_ReadDataRemelt(int layernumber, int id, int nx, int MyYSlices, int
                 for (int j = (i + 1); j < NumberOfSolidificationEvents_Host(n); j++) {
                     if (LayerTimeTempHistory_Host(n, i, 0) > LayerTimeTempHistory_Host(n, j, 0)) {
                         // Swap these two points - melting event "j" happens before event "i"
-                        float OldMeltVal = LayerTimeTempHistory_Host(n, i, 0);
```
streeve, line 1578: Shouldn't this still be float?

streeve (approved):
In reading through this, it looks consistent now for `double`. But it does look like there are a lot of hidden conversions to `float` (and a number of c-style casts to `int`). Up to you if you want a separate PR for those

PR #151, opened 2022-12-08
Optional binary reading of temperature data

MattRolchigo:
- Should not be merged until #150 is finished, as this will need to be rebased
- Instead of having the option to read binary input data given in the temperature instructions file (potentially mixing up which ones use ASCII and which ones use binary), John and I discussed defining our binary string of double precision values as a special file type (.exaca), and if a temperature input file has that extension, automatically parse it as binary? And otherwise (.csv, .txt, etc) treat as ASCII?


MattRolchigo:
Pretty sure the CI failure is related to how float and double were used interchangeably by mistake, which will be fixed in #150

src/CAparsefiles.cpp
```diff
@@ -367,3 +372,156 @@ std::string parseCoordinatePair(std::string line, int val) {
     SplitStr = std::regex_replace(SplitStr, r, "");
     return SplitStr;
 }
+
+// Read x, y, z coordinates in tempfile_thislayer (temperature file in either an ASCII or binary format) and return the
+// min and max values
+std::array<double, 6> parseCoordinateMinMax(std::string tempfile_thislayer, bool BinaryInputData) {
```
streeve, line 378: parseTemperatureCoordinateMinMax?

src/CAparsefiles.cpp
```diff
@@ -367,3 +372,156 @@ std::string parseCoordinatePair(std::string line, int val) {
...
+
+    std::array<double, 6> XYZMinMax;
+    std::ifstream TemperatureFilestream;
+    TemperatureFilestream.open(tempfile_thislayer);
+
+    if (!(BinaryInputData)) {
+        // Read the header line data
+        // Make sure the first line contains all required column names: x, y, z, tm, tl, cr
+        // Check 2 mixes of lower and uppercase letters/similar possible column names just in case
```
streeve, line 387: Comment seems out of date

src/CAparsefiles.cpp
```diff
@@ -367,3 +372,156 @@ std::string parseCoordinatePair(std::string line, int val) {
...
+        std::string HeaderLine;
+        getline(TemperatureFilestream, HeaderLine);
+        checkForHeaderValues(HeaderLine);
+    }
+
+    // Units are assumed to be in meters, meters, seconds, seconds, and K/second
+    std::vector<double> XCoordinates(1000000), YCoordinates(1000000), ZCoordinates(1000000);
```
streeve, line 394: Can we at least set 1M as a local variable above or input with default?

src/CAinitialize.cpp
```diff
@@ -569,76 +569,31 @@ void FindXYZBounds(std::string SimulationType, int id, double &deltax, int &nx,
...
-            TemperatureFile.close();
-            // Min/max x, y, and z coordinates from this layer's data
-            XMin_ThisLayer = *min_element(XCoordinates.begin(), XCoordinates.end());
-            YMin_ThisLayer = *min_element(YCoordinates.begin(), YCoordinates.end());
-            ZMin_ThisLayer = *min_element(ZCoordinates.begin(), ZCoordinates.end());
-            XMax_ThisLayer = *max_element(XCoordinates.begin(), XCoordinates.end());
-            YMax_ThisLayer = *max_element(YCoordinates.begin(), YCoordinates.end());
-            ZMax_ThisLayer = *max_element(ZCoordinates.begin(), ZCoordinates.end());
+            // Get min and max x coordinates in this file, which can be a binary or ASCII input file
+            // index 0 = min x, index 1 = max x, index 2 = min y, index 3 = max y, index 4 = min z, index 5 = maxz
```
streeve, line 573: `// { Xmin, Xmax, Ymin, Ymax, Zmin, Zmax }`

streeve (commented):
Just minor stuff

src/CAparsefiles.cpp
MattRolchigo: Yeah, we have two options for each of the last two columns, but there shouldn't be any capital letters in the header line

src/CAinitialize.cpp
```diff
@@ -528,6 +528,17 @@ void NeighborListInit(NList &NeighborX, NList &NeighborY, NList &NeighborZ) {
     NeighborZ = {0, 0, 0, 1, -1, 1, 1, -1, -1, 1, -1, 1, 1, -1, -1, 0, 0, 0, 0, 0, 1, -1, 1, 1, -1, -1};
 }
 
+// Check if the temperature data is in ASCII or binary format
+bool checkTemperatureFileFormat(std::string tempfile_thislayer) {
+    bool BinaryInputData;
+    std::size_t found = tempfile_thislayer.find(".exaca");
```
streeve, line 534: Should update the readme for this

src/CAinitialize.cpp
```diff
@@ -569,76 +580,33 @@ void FindXYZBounds(std::string SimulationType, int id, double &deltax, int &nx,
...
-                YMax = YMax_ThisLayer;
-            if (ZMin_ThisLayer < ZMin)
-                ZMin = ZMin_ThisLayer;
-            if (ZMax_ThisLayer > ZMax)
-                ZMax = ZMax_ThisLayer;
-            ZMinLayer[LayerReadCount - 1] = ZMin_ThisLayer;
-            ZMaxLayer[LayerReadCount - 1] = ZMax_ThisLayer;
+            XYZMinMax[4] += deltax * LayerHeight * (LayerReadCount - 1);
+            XYZMinMax[5] += deltax * LayerHeight * (LayerReadCount - 1);
+            if (XYZMinMax[0] < XMin)
```
streeve, line 594: The array is global, right? If so, can we rename?

src/CAinitialize.hpp
```diff
@@ -10,6 +10,7 @@
 
 #include <Kokkos_Core.hpp>
 
+#include <iomanip>
```
streeve, line 13: Don't see where you're using this

unit_test/tstInit.cpp
```diff
@@ -602,11 +708,16 @@ TEST(TEST_CATEGORY, activedomainsizecalc) {
     testcalcLocalActiveDomainSize();
 }
 TEST(TEST_CATEGORY, temperature_init_test) {
+    // reading temperature files to obtain xyz bounds, using binary/non-binary format
+    testFindXYZBounds(false);
+    testFindXYZBounds(true);
     // reading one file of temperature data - performed in the same manner with and without remelting
```
streeve, line 714: Maybe just loop over true/false for both inputs?

streeve (commented):
Just small stuff

examples/README.md
```diff
@@ -5,9 +5,10 @@ ExaCA currently can model three types of problems, two of which have the option
...
     * If an x,y,z coordinate melted and solidified multiple times, it should appear in the file multiple times on separate lines. The order of the lines do not matter, except that the header line must be before any data.
     * The top surface (the largest Z coordinate in a file) is assumed to be flat. Additionally, if multiple temperature files are being used (for example, a scan pattern consisting of 10 layers of repeating even and odd file data), the Z coordinate corresponding to this flat top surface should be the same for all files.
+    * Alternatively, if a time-temperature history file has the extension `.exaca`, it will be parsed as a binary string. The binary form for these files does not contain commas, newlines, nor a header, but consists of sequential x,y,z,tm,tl,cr,x,y,z,tm,tl,cr... data as double precision values (little endian). This is often a significantly smaller file size than the standard format, and will be faster to read during initialization.
```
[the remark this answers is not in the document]
MattRolchigo: .exaca files aren't written out during ExaCA simulations, it's just the input file format that's read and parsed by ExaCA

src/CAinitialize.cpp
MattRolchigo: Global as in the coordinates are relative to the full multilayer domain? Yes. But I don't use local/global labels for any of the coordinate based variables (XMin, XMax, etc) since the global coordinate system is the only one used (the local sizes and subdomain offsets are only used for the dimensionless positions/sizes, i.e., ny and MyYSlices)

streeve: Then I guess my confusion is why we're checking Xmin etc. against XYZMinMax if we just read in the array

unit_test/tstInit.cpp
MattRolchigo: I updated the test, not sure why it isn't showing up as "outdated"

examples/README.md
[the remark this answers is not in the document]
streeve: I would expect `.exaca` to be a checkpoint/restart/state file. `.amfoamtemp` or something is really what it is

unit_test/tstInit.cpp
streeve: I don't see any changes regarding this comment

src/CAinitialize.cpp
MattRolchigo: This is inside of a loop over all temperature files - parseTemperatureCoordinateMinMax only returns the XYZMinMax coordinates for whatever file was passed to it. I'll rename XYZMinMax to XYZMinMax_ThisLayer in CAinitialize.cpp for clarity

unit_test/tstInit.cpp
streeve: Oh now I see - passing the bool to the function seems fine, I just meant in this top-level function you could go through all the bools in a double loop

unit_test/tstInit.cpp
```diff
@@ -691,14 +685,18 @@ TEST(TEST_CATEGORY, activedomainsizecalc) {
...
-    // reading temperature data when two files are present - reading either one (true) or both (false)
-    testReadTemperatureData(2, true, false);
-    testReadTemperatureData(2, false, false);
-    // read both temperature data files in binary string format
-    testReadTemperatureData(2, true, true);
+    testFindXYZBounds(false);
+    testFindXYZBounds(true);
+    // Multiple permutations of inputs: NumberOfLayers, LayerwiseTempRead, TestBinaryInputRead
+    // reading temperature data is performed in the same manner with and without remelting
+    std::vector<int> NumberOfLayers_vals = {1, 2, 2, 2};
```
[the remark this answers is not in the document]
MattRolchigo: LayerwiseTempRead is only relevant when NumberOfLayers > 1, and if the binary file read works for reading 2 files, it should work for 1 file as well. I wasn't sure if it would be worth it to run tests for additional combinations of options to look for fringe edge cases here

PR #154, opened 2022-12-12
Avoid out of bounds error during temperature interpolation

MattRolchigo:
Fixed incorrect use of std::numeric_limits and truncated interpolation domain to avoid interpolating regions outside the simulation domain

### DIFF
--- a/src/CAinitialize.cpp
+++ b/src/CAinitialize.cpp
@@ -517,9 +517,9 @@ void FindXYZBounds(std::string SimulationType, int id, double &deltax, int &nx, 
         XMin = std::numeric_limits<double>::max();
         YMin = std::numeric_limits<double>::max();
         ZMin = std::numeric_limits<double>::max();
-        XMax = std::numeric_limits<double>::min();
-        YMax = std::numeric_limits<double>::min();
-        ZMax = std::numeric_limits<double>::min();
+        XMax = std::numeric_limits<double>::lowest();
+        YMax = std::numeric_limits<double>::lowest();
+        ZMax = std::numeric_limits<double>::lowest();
 
         // Read the first temperature file, first line to determine if the "new" OpenFOAM output format (with a 1 line
         // header) is used, or whether the "old" OpenFOAM header (which contains information like the X/Y/Z bounds of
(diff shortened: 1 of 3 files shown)


src/runCA.cpp
```diff
@@ -151,7 +151,7 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
         TempInit_ReadDataNoRemelt(id, nx, MyYSlices, MyYOffset, deltax, HTtoCAratio, deltat, nz, LocalDomainSize,
```
streeve, line 151: id, nx, ny?

src/CAinitialize.cpp
```diff
@@ -1201,6 +1201,9 @@ void TempInit_ReadDataNoRemelt(int id, int &nx, int &MyYSlices, int &MyYOffset,
     // from HT_deltax to deltax
     int LowerYBound = MyYOffset - (MyYOffset % HTtoCAratio);
     int UpperYBound = MyYOffset + MyYSlices - 1 + HTtoCAratio - ((MyYOffset + MyYSlices - 1) % HTtoCAratio);
+    // Make sure that upper Y bound doesn't extend beyond simulation domain
+    if (UpperYBound >= ny)
+        UpperYBound = ny - 1;
```
streeve, line 1206: Is it better to allow equal and error if it's larger? Either way is fine with me

src/CAinitialize.cpp
```diff
@@ -517,9 +517,9 @@ void FindXYZBounds(std::string SimulationType, int id, double &deltax, int &nx,
         XMin = std::numeric_limits<double>::max();
         YMin = std::numeric_limits<double>::max();
         ZMin = std::numeric_limits<double>::max();
-        XMax = std::numeric_limits<double>::min();
-        YMax = std::numeric_limits<double>::min();
-        ZMax = std::numeric_limits<double>::min();
+        XMax = std::numeric_limits<double>::lowest();
```
streeve, line 520: TIL this distinction

src/runCA.cpp
MattRolchigo: I thought about that, but I've started always adding new inputs to the end of the function definitions, to avoid issues where adding a new input to the first line causes all of the other lines to rearrange and show up as a change set. Once we have a grid object we don't have to worry about this anymore

src/CAinitialize.cpp
MattRolchigo: Well the domain itself goes from 0 to ny-1 inclusive, so if an MPI rank has an upper bound beyond that, it should be truncated at ny-1. This is a little different than other places in the code where the range doesn't include the upper bound value itself

streeve (approved):

PR #155, opened 2022-12-12
Add version/commit info

streeve:
Prints at configure and run time


src/CAprint.hpp
```diff
@@ -67,6 +67,9 @@ void PrintIntermediateExaCAState(int IntermediateFileCounter, int layernumber, s
                                  ViewI3D_H GrainID_WholeDomain, ViewI3D_H CellType_WholeDomain, ViewF_H GrainUnitVector,
                                  int NGrainOrientations, double deltax, double XMin, double YMin, double ZMin,
                                  bool PrintBinary);
+std::string version();
```
MattRolchigo, line 70: So you get the version and the commit hash here, but you don't do anything with them? Or are they printed to the console when you call these functions?

src/main.cpp
```diff
@@ -25,11 +25,11 @@ int main(int argc, char *argv[]) {
         // Get individual process ID
         MPI_Comm_rank(MPI_COMM_WORLD, &id);
 
-        if (id == 0)
+        if (id == 0) {
+            std::cout << "ExaCA version: " << version() << " \nExaCA commit:  " << gitCommitHash() << std::endl;
```
MattRolchigo, line 29: We should have this also appear in the log file created at the end of the simulation

MattRolchigo (changes requested):

src/CAprint.hpp
streeve: You can do anything you want with the strings, yeah

src/main.cpp
streeve: Done

MattRolchigo (approved):

PR #156, opened 2022-12-14
Option to use json format for ExaCA main input files

MattRolchigo:


streeve (changes requested):
Looks like it's failing because a json file calls the old input

.github/workflows/CI.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             backend: 'OPENMP'
             cmake_build_type: 'Debug'
             kokkos_ver: '3.4.01'
-            json: 'OFF'
```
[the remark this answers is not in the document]
MattRolchigo: In that case, should the CI only be running with the .txt version of the input problems?

src/CAinitialize.cpp
```diff
@@ -471,6 +496,272 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
     }
 }
 
+// Read ExaCA input file - new (json) file format
+#ifdef ExaCA_ENABLE_JSON
+void InputReadFromFile_json(int id, std::string InputFile, std::string &SimulationType, double &deltax, double &NMax,
```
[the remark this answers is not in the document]
MattRolchigo: So do you think that the "new" versions of functions should just be given the old name, and the "old" versions named things like `InputReadFromFile_old`?

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,50 @@
...
+      "S0": 25
+      "powderDensity": 8000
+   },
+   "Printing": {
+       "pathToOutput": "./",
+       "outputFile": "TestProblemAMB",
+       "printBinary": false,
+       "printExaConstitDefault": false,
+       "printExaConstitSize": 0,
+       "PrintFieldsInit": {
```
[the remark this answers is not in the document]
MattRolchigo: I actually tried that first, but in that case you have to check each position in the list against each possible input, and I hadn't realized that using .count() was a thing so it was harder than it should've been

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,50 @@
...
+       "deltat": 0.3,
+       "numberOfLayers": 4,
+       "layerOffset": 3
+   },
+   "Nucleation": {
+      "NMax": 0,
+      "dTN": 5,
+      "dTsigma": 0.5
+   },
+   "TemperatureData": {
```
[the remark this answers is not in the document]
MattRolchigo: Yes because for the problems that don't use a temperature data file (C and S), the thermal gradient and cooling rate would be given here. Unless we made separate temperature input files for those problem types analogous to the files like `examples/Temperatures/T_AMBenchMultilayer.json`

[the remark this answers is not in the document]
MattRolchigo: And if we were going to do that, I'd rather just get rid of the temperature input files altogether and just have that information in the main input file

.github/workflows/CI.yml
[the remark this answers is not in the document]
streeve: Finally fixed this

streeve (commented):
CI errors now seem to be about the input file itself

examples/Inp_AMBenchMultilayer.json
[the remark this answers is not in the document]
MattRolchigo: As much as having the list in the input file is kind of clunky and adds extra lines, I actually think I like it because you can see what all the options are and easily toggle them on or off. As opposed to having to go back into the README and remembering what the views were called and which ones can be printed

src/CAparsefiles.cpp
```diff
@@ -218,6 +222,91 @@ void checkFileNotEmpty(std::string testfilename) {
...
+                                 double &HT_deltax, std::vector<std::string> &temp_paths, bool &RemeltingYN,
+                                 bool &LayerwiseTempRead) {
+
+    // Check that file exists and contains data
+    checkFileExists(TFieldInstructions, id);
+    checkFileNotEmpty(TFieldInstructions);
+    std::ifstream TemperatureData(TFieldInstructions);
+    nlohmann::json tinputdata = nlohmann::json::parse(TemperatureData);
+    // If this input was not given, default to using CA cell size as the assumed temperature data resolution
+    if (tinputdata["HTdeltax"] != nullptr) {
```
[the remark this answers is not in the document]
MattRolchigo: I replaces all of the nullptr comparisons with the "contains" member function

examples/Inp_AMBenchMultilayer.json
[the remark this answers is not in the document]
MattRolchigo: I think it's a lot easier to follow without the temperatures input files

MattRolchigo:
@streeve It looks like it's still failing on the json testing option

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,39 @@
...
+   },
+   "Printing": {
+       "pathToOutput": "./",
+       "outputFile": "TestProblemAMB",
+       "printBinary": false,
+       "printExaConstitDefault": false,
+       "printExaConstitSize": 0,
+       "PrintFieldsInit": [],
+       "PrintFieldsFinal": ["GrainID", "LayerID", "GrainMisorientation"],
+       "PrintIntermediateOutput": {
```
streeve, line 34: Should this be `PrintFieldsIntermediate`? And then (in a separate PR) allow choosing which fields to print just like the others? Potentially even removing the init/final options and relying entirely on the frequency of printing

src/CAinterfacialresponse.hpp
```diff
@@ -39,65 +39,66 @@ struct InterfacialResponseFunction {
...
+                throw std::runtime_error(error);
             }
-            MaterialData.close();
-            FreezingRange = getInputDouble(MaterialInputsRead[0]);
-            A = getInputDouble(MaterialInputsRead[1]);
-            B = getInputDouble(MaterialInputsRead[2]);
-            C = getInputDouble(MaterialInputsRead[3]);
-            D = getInputDouble(MaterialInputsRead[4]);
         }
-        normalize(deltat, deltax);
```
streeve, line 95: Was this a bug?

src/CAinitialize.cpp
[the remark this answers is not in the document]
streeve: That would let us avoid another rename when we remove the `old` versions

src/runCA.cpp
```diff
@@ -33,16 +33,34 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
...
+        InputReadFromFile_json(id, InputFile, SimulationType, deltax, NMax, dTN, dTsigma, OutputFile,
+                               GrainOrientationFile, TempFilesInSeries, temp_paths, HT_deltax, RemeltingYN, deltat,
+                               NumberOfLayers, LayerHeight, MaterialFileName, SubstrateFileName, SubstrateGrainSpacing,
+                               UseSubstrateFile, G, R, nx, ny, nz, FractSurfaceSitesActive, PathToOutput, PrintDebug,
+                               PrintMisorientation, PrintFinalUndercoolingVals, PrintFullOutput, NSpotsX, NSpotsY,
+                               SpotOffset, SpotRadius, PrintTimeSeries, TimeSeriesInc, PrintIdleTimeSeriesFrames,
+                               PrintDefaultRVE, RNGSeed, BaseplateThroughPowder, PowderDensity, RVESize,
+                               LayerwiseTempRead, PrintBinary);
+#else
+        throw std::runtime_error("Error: attempted to parse json input file without ExaCA_ENABLE_JSON=ON");
```
streeve, line 51: Shouldn't this already be accounted for in `checkInputFileFormat`?

unit_test/tstInit.cpp
```diff
@@ -570,7 +644,12 @@ void testgetTempCoords() {
 // RUN TESTS
 //---------------------------------------------------------------------------//
 TEST(TEST_CATEGORY, fileread_test) {
-    testInputReadFromFile();
+    // test for reading new input file format (true) and old (false)
+    // FIXME: remove test for old input parsing in future release
+    //  testInputReadFromFile(false);
```
streeve, line 649: Why is this commented?

streeve (commented):
Looks almost ready

unit_test/tstInit.cpp
```diff
@@ -96,29 +97,12 @@ void WriteTestDataR_json(std::string InputFilename) {
...
-    TestDataFile << "      }," << std::endl;
-    TestDataFile << "      \"PrintFieldsFinal\": {" << std::endl;
-    TestDataFile << "         \"GrainID\": false," << std::endl;
-    TestDataFile << "         \"LayerID\": false," << std::endl;
-    TestDataFile << "         \"UndercoolingCurrent\": true," << std::endl;
-    TestDataFile << "         \"GrainMisorientation\": true" << std::endl;
-    TestDataFile << "      }" << std::endl;
+    TestDataFile << "      \"printExaConstitDefault\": true," << std::endl;
+    TestDataFile << "      \"PrintFieldsInit\": []," << std::endl;
+    TestDataFile << "      \"PrintFieldsFinal\": [\"UndercoolingCurrent\",\"GrainMisorientation\"]" << std::endl;
```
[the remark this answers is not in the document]
MattRolchigo: Yeah, we can have this test run for a "debug" case (all initial fields printed, final undercooling printed, grain misorientation and grainID/layerID printed) and for a "standard" case (grain misorientation and grainID/layerID printed, default ExaConstit output printed)

examples/Inp_AMBenchMultilayer.json
MattRolchigo: The intermediate output is distinct from the init and final print options because it prints something other than just one field: it prints grain misorientation values for all cells that are non-liquid at that time step, and values of -1 for all liquid cells. Which is the reason it is named something slightly different. But I agree though that printing intermediate values of the other fields could be useful as well, the print options needs a few PR updates actually. They still don't work for runs with remelting, there's only 3 permutations of init fields that can be printed- "no debug", "light debug", and "full debug", and storing the options in a struct will simplify the code a lot

src/CAinterfacialresponse.hpp
MattRolchigo: No, I just moved the call to normalize into the constructor itself (it is now on line 55), rather than have it called separately in each of `parseMaterialJson` and `parseMaterialOld`

src/runCA.cpp
MattRolchigo: Yeah, that was just a redundant check I forgot to remove

unit_test/tstInit.cpp
MattRolchigo: Because I'm dumb

streeve (commented):
Few small things. Also, the capitalization of the new inputs is inconsistent, but I don't have a strong preference for which you choose. Are you going to add the analysis parsing here as well?

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,39 @@
...
+       "numberOfLayers": 4,
+       "layerOffset": 3
+   },
+   "Nucleation": {
+      "NMax": 0,
+      "dTN": 5,
+      "dTsigma": 0.5
+   },
+   "TemperatureData": {
+       "HTdeltax": 5,
```
streeve, line 18: Good opportunity to make some of these variables more readable (HeatTransferDeltaX)

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,39 @@
...
+       "layerwiseTempRead": false,
+       "TemperatureFiles": ["examples/Temperatures/1CA-FOAM-fluidflow.txt","examples/Temperatures/2CA-FOAM-fluidflow.txt"]
+   },
+   "Substrate": {
+      "S0": 25,
```
streeve, line 23: No idea what this is

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,39 @@
...
+   "MaterialFileName": "Inconel625.json",
+   "GrainOrientationFile": "GrainOrientationVectors.csv",
+   "RNGSeed": 0.0,
+   "Domain": {
+       "deltax": 5,
+       "deltat": 0.3,
+       "numberOfLayers": 4,
+       "layerOffset": 3
+   },
+   "Nucleation": {
```
streeve, line 12: All of these are a bit unclear

examples/Inp_AMBenchMultilayer.json
```diff
@@ -0,0 +1,39 @@
...
+   },
+   "Substrate": {
+      "S0": 25,
+      "powderDensity": 8000
+   },
+   "Printing": {
+       "pathToOutput": "./",
+       "outputFile": "TestProblemAMB",
+       "printBinary": false,
+       "printExaConstitDefault": false,
```
streeve, line 30: Worth a separate ExaConstit subsection?

MattRolchigo: Maybe just "PrintExaConstitSize" where setting the size to 0 or not including the input tells the code not to print it (which is also what PrintIntermediateOutput does if Frequency = 0). Unlike PrintIntermediateOutput which has 2 inputs (the frequency of output and whether or not to print data between laser passes), this only has size as an input, so it didn't seem worthy of its own section.

streeve: Not sure how you use one input for both default size and to turn it off. This would add an easy way to turn it off (without removing entirely)
```
"ExaConstitRVE"
{
  "Print": false,
  "Default": false,
  "LengthMicrons": 10,
}
```

MattRolchigo: There wouldn't be a default size in this case, a size would just always be required and if no size is given, nothing is printed. The default size is kind of arbitrary considering that we picked it based on what we wanted for the AMBench runs but wouldn't be all that useful for anyone else. It just seems like such a rarely used feature for most users to necessitate its own section

streeve: Fair enough

MattRolchigo:
Title updated to reflect that the analysis input file formatting will be in a separate PR

streeve (commented):
Sorry I wasn't clear in the last two places. Other than those and updating the readme I think it's ready

examples/Inp_AMBenchMultilayer.json
```diff
@@ -2,17 +2,17 @@
    "SimulationType": "R",
    "MaterialFileName": "Inconel625.json",
    "GrainOrientationFile": "GrainOrientationVectors.csv",
-   "RNGSeed": 0.0,
+   "RandomSeed": 0,
```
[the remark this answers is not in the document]
MattRolchigo: Yeah I realized I'm being inconsistent with it as-is, the random seed is stored as a double in the code, but when it's used in a function, that function requires the seed to be an int so it must just automatically be getting converted. We should probably should make a new issue about that...

examples/Inp_AMBenchMultilayer.json
```diff
@@ -2,17 +2,17 @@
...
-       "Deltax": 5,
-       "Deltat": 0.3,
+       "CellSize": 5,
+       "TimeStep": 0.3,
        "NumberOfLayers": 4,
        "LayerOffset": 3
    },
    "Nucleation": {
       "Density": 0,
-      "MeanUndercooling": 5,
```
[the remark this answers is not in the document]
MattRolchigo: Probably not, now that I think about it... I was just wanting to keep it consistent with StDev, but at least one of the two should have Undercooling in there so that we know what the distribution actually is

streeve (approved):
Looks good to me

PR #160, opened 2023-01-04
Check for max MPI decomposition

### LINKED ISSUE #24: Code should throw error when problem size is incompatible with resource set
Throw an error if a problem is run with a number of resources that cannot be utilized to solve it. For example, a domain 200 cells in the Y direction with a 1D domain decomposition scheme cannot be run on 220 MPI ranks/CPUs as there is no way to distribute the work properly. Otherwise, this edge case will result in either incorrect results or an infinite loop in the code

streeve:
Closes #24


MattRolchigo (approved):
I feel like we had an issue like this open at one point and I thought I had addressed it, but maybe I dreamed that

streeve:
> I feel like we had an issue like this open at one point and I thought I had addressed it, but maybe I dreamed that

Linked in description

PR #163, opened 2023-01-12
Reorganize PrintExaCALog/ensure all information is printed

MattRolchigo:
Previously, the ExaCA commit hash and the paths/names of the temperature files were printed to std::cout rather than the log. `PrintExaCALog` was also rearranged slightly


streeve (approved):
Looks good except that this broke the analysis parsing

streeve (changes requested):
I think we should strongly consider printing the log to JSON instead

analysis/src/GAutils.cpp
```diff
@@ -59,8 +59,74 @@ int FindTopOrBottom(ViewI3D_H LayerID, int XLow, int XHigh, int YLow, int YHigh,
...
+    getline(InputDataStream, testline);
+    std::size_t found_version = testline.find("ExaCA version");
+    if (found_version == std::string::npos)
+        NewLogFormatYN = false;
+    else
+        NewLogFormatYN = true;
+    return NewLogFormatYN;
+}
+
 void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax, int &NumberOfLayers,
```
streeve, line 80: Add comment about use of main parsing options

streeve (approved):

PR #165, opened 2023-01-19
Build analysis as separate library with unit test support

streeve:
First commit: create analysis library
Second: make adding analysis unit tests easier


MattRolchigo (commented):
So to be clear, the top level unit tests directory is the same as before, and there's a new unit tests subdirectory within the analysis directory for unit tests of functions within GAutils.cpp, etc? Worth separating the analysis unit tests into with/without kokkos as well, or just put them all into what is currently named tstEmpty.hpp?

streeve:
> So to be clear, the top level unit tests directory is the same as before, and there's a new unit tests subdirectory within the analysis directory for unit tests of functions within GAutils.cpp, etc? 
Yes

> Worth separating the analysis unit tests into with/without kokkos as well, or just put them all into what is currently named tstEmpty.hpp?
I'd say separate however is helpful (and rename that however you want)

MattRolchigo:
Sounds good to me

MattRolchigo (approved):

PR #166, opened 2023-01-23
Use default Euler angles file, MTEX printing routines

MattRolchigo:
The scripts needed in Matlab to plot cross-section IPF-colored data or pole figure data are much more simple when using the Euler angle file, and this doesn't cause any backward compatibility issues in ExaCA code itself. It's a lot easier to stick to one plotting format for when I add the matlab plotting scripts to the repository


streeve (approved):
Isn't a compatibility issue because mtex can read either one?

MattRolchigo:
Yeah, you need a different script to read either one but everything works a lot cleaner when printing the Euler angles directly

PR #167, opened 2023-01-23
Helper functions and analysis print routine rewrite

MattRolchigo:
* Separate calculations into separate helper functions (previously parts of print subroutines in GAprint.cpp)
* Printing grain-wise statistics to a single file, edited by multiple print routines, rather than to individual files
* Keep old calculated quantities that are no longer used with warnings that they will not be computed in a future release


analysis/unit_test/tstKokkosUtils.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include <gtest/gtest.h>
+
+#include <string>
+#include <vector>
+
+namespace Test {
+//---------------------------------------------------------------------------//
+// GAutils tests with Kokkos
+//---------------------------------------------------------------------------//
+void testgetRepresentativeRegionGrainIDs() {
```
streeve, line 19: Can you add top level test comments? Just not initially sure what the setup/intent is

analysis/src/GAprint.cpp
```diff
@@ -7,20 +7,108 @@
...
+    GrainStatistics.open(GrainStatistics_fname);
+    GrainStatistics << "GrainID:";
+    for (int n = 0; n < NumberOfGrains - 1; n++) {
+        GrainStatistics << UniqueGrainIDVector[n] << ",";
+    }
+    GrainStatistics << UniqueGrainIDVector[NumberOfGrains - 1] << std::endl;
+
+    std::string QoIs_fname = OutputFileName + "_grains.csv";
+    QoIs.open(QoIs_fname);
+    std::cout << "Note: lines that start with *** are information that will no longer be collected/printed to the "
```
streeve, line 27: Why not just add a line with "Warning: the analysis is deprecated and will be removed" before each old section? Not sure about motivation for the pseudo comment *** here

analysis/src/GAprint.cpp
```diff
@@ -45,240 +131,207 @@ void PrintMisorientationData(bool *AnalysisTypes, std::string BaseFileName, int
...
-    double AvgVolPerGrain_Norm = AvgVolPerGrain * deltax * deltax * pow(10, 12);
-    std::cout << "-- There are " << NumberOfGrains << " grains in this volume, and the mean grain volume is "
-              << AvgVolPerGrain_Norm << " cubic microns" << std::endl;
-    double VolFractNucGrains = DivideCast<double>(NucleatedGrainCells, DomainVol);
-    std::cout << "-- The volume fraction consisting of nucleated grains is " << VolFractNucGrains << std::endl;
+}
+
+// Print average aspect ratio in the build to the average of the transverse directions, and the ratio for each
+// individual grain
+void PrintBuiltTransAspectRatioData(std::ofstream &QoIs, std::ofstream &GrainStatistics, std::vector<int> GrainExtentX,
```
streeve, line 181: Would rather keep `printBuiltTransverseAspectRatio`. I don't think every function needs `Data` at the end and would rather a long name than shortening something like `Trans`

analysis/src/GAprint.cpp
```diff
@@ -45,240 +131,207 @@ void PrintMisorientationData(bool *AnalysisTypes, std::string BaseFileName, int
...
-        for (int i = XMin; i <= XMax; i++) {
-            for (int j = YMin; j <= YMax; j++) {
-                AllGrainIDs[Counter] = GrainID(k, i, j);
-                Counter++;
-                if (GrainID(k, i, j) < 0)
-                    NucleatedGrainCells++;
-            }
+// Print the average grain size, and the size of each grain from the vector (if PrintStats = true). Default is 3D size
+// (volumes)
+void PrintSizeData(bool PrintStats, std::ofstream &QoIs, std::ofstream &GrainStatistics,
```
streeve, line 144: `printGrainSize`?

streeve (commented):
Haven't looked close yet, but there's a conflict anyway

analysis/unit_test/tstKokkosUtils.hpp
MattRolchigo: Added a comment before each unit in analysis/unit_test briefly describing what each test does

analysis/src/GAutils.cpp
```diff
@@ -783,13 +783,77 @@ void ParseAnalysisFile(std::string AnalysisFile, std::string RotationFilename, i
...
+                        else if (Direction == "Y")
+                            GrainCoordinate[count] = j;
+                        else if (Direction == "Z")
+                            GrainCoordinate[count] = k;
+                        count++;
+                    }
+                }
+            }
+        }
+        int MinCoord = *min_element(GrainCoordinate.begin(), GrainCoordinate.end());
```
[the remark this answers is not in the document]
MattRolchigo: Consistency with what I did in src/CAparsefiles.cpp `parseTemperatureCoordinateMinMax` I guess

analysis/src/GAutils.cpp
```diff
@@ -919,3 +969,102 @@ ViewI_H createOrientationHistogram(int NumberOfOrientations, std::vector<int> Gr
     }
     return GOHistogram;
 }
+
+// Given an input vector of integer Grain ID values "GrainIDVector", and an input vector of the
+// unique Grain ID values "UniqueGrainIDVector" (of size "NumberOfGrains"), return a third vector "GrainSizeVector"
+// listing the size of each of the "NumberOfGrains" grains (scaled by the cell size deltax and depending on whether the
+// region is 1D, 2D, or 3D)
+std::vector<float> getGrainSizes(const std::vector<int> GrainIDVector, const std::vector<int> UniqueGrainIDVector,
+                                 const int NumberOfGrains, double deltax, std::string RegionType) {
```
streeve, line 978: Same here - can't you assume the size of `UniqueGrainIDVector` is the number of grains?

streeve (changes requested):
Last changes I think

analysis/src/GAutils.cpp
MattRolchigo: You mean not passing NumberOfGrains as an argument since it's the same as the size of UniqueGrainIDVector?

analysis/src/GAutils.cpp
```diff
@@ -783,13 +783,159 @@ void ParseAnalysisFile(std::string AnalysisFile, std::string RotationFilename, i
...
-// sorted from lowest to highest
-std::vector<int> FindUniqueGrains(const std::vector<int> GrainIDVector) {
+// sorted from lowest to highest. Store the number of grains
+std::vector<int> getUniqueGrains(const std::vector<int> GrainIDVector, int &NumberOfGrains) {
     std::vector<int> UniqueGrainIDVector = GrainIDVector;
     std::sort(UniqueGrainIDVector.begin(), UniqueGrainIDVector.end());
     std::vector<int>::iterator it;
     it = std::unique(UniqueGrainIDVector.begin(), UniqueGrainIDVector.end());
     UniqueGrainIDVector.resize(std::distance(UniqueGrainIDVector.begin(), it));
+    NumberOfGrains = UniqueGrainIDVector.size();
```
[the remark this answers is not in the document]
streeve: Confused here since it is inside `getUniqueGrains`, but no strong feelings on the separate variable/size

analysis/src/GAutils.cpp
streeve: Yes, but it sounds like you'd prefer to

analysis/src/GAutils.cpp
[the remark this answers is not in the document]
MattRolchigo: `getUniqueGrains` also has the line `NumberOfGrains = UniqueGrainIDVector.size();` inside of it, `NumberOfGrains` is passed uninitialized by reference into `getUniqueGrains` to avoid setting it to the wrong thing outside the function where the unique grain IDs are obtained

streeve (approved):

PR #169, opened 2023-01-27
Print and parse log files with json format

MattRolchigo:
Prints log files as json if json is enabled (otherwise using old format). Maintains capability to read log files of previous formats

### DIFF
--- a/analysis/src/GAutils.cpp
+++ b/analysis/src/GAutils.cpp
@@ -355,25 +355,6 @@ void ReadIgnoreField(std::ifstream &InputDataStream, int nx, int ny, int nz) {
         getline(InputDataStream, line);
 }
 
-// Check if log file format is consistent with compilation option, swap form if necessary
-std::string CheckLogFilename(std::string LogFile) {
-    std::string LogFileUpdate;
-    bool LogFormatConsistent = checkFileExists(LogFile, 0, false);
-    if (LogFormatConsistent)
-        LogFileUpdate = LogFile;
-    else {
-        // Use the other possible version of the log file
-        std::size_t extension_pos = LogFile.find_last_of(".");
-        std::string LogFileNamePrefix = LogFile.substr(0, extension_pos);
-        std::string LogFileExt = LogFile.substr(extension_pos, std::string::npos);
-        if (LogFileExt == ".json")
-            LogFileUpdate = LogFileNamePrefix + ".log";
-        else if (LogFileExt == ".log")
-            LogFileUpdate = LogFileNamePrefix + ".json";
-    }
-    return LogFileUpdate;
-}
-
 // Read the analysis file to determine the file names/paths for the microstructure and the orientations
 void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string &MicrostructureFile,
                     std::string &RotationFilename, std::string &OutputFileName, std::string &EulerAnglesFilename,
@@ -387,16 +368,20 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string 
         throw std::runtime_error("Error: Cannot find ExaCA analysis file");
     skipLines(Analysis, "*****");
 
-    std::vector<std::string> NamesOfFilesRequired = {"name of log", "name of microstructure (.vtk) file",
+    std::vector<std::string> NamesOfFilesRequired = {"name of microstructure (.vtk) file",
                                                      "name of data files of output resulting from this analysis"};
     std::vector<std::string> NamesOfFilesOptional = {
-        "file of grain orientations (rotation matrix form)", "file of grain orientations (Bunge Euler angle form ZXZ)",
+        "Path to/name of log file", "file of grain orientations (rotation matrix form)",
+        "file of grain orientations (Bunge Euler angle form ZXZ)",
         "file corresponding to grain orientation IPF-Z colors as fractional RGB values"};
+    std::vector<std::string> NamesOfFilesDeprecated = {"Path to/name of log (.log) file"};
     // If orientation file names are given here, store these values but override if given in the log file with a warning
     int NumRequiredInputs = NamesOfFilesRequired.size();
     int NumOptionalInputs = NamesOfFilesOptional.size();
+    int NumDeprecatedInputs = NamesOfFilesDeprecated.size();
     std::vector<std::string> RequiredFilesRead(NumRequiredInputs);
     std::vector<std::string> OptionalFilesRead(NumOptionalInputs);
+    std::vector<std::string> DeprecatedFilesRead(NumDeprecatedInputs);
 
     for (int i = 0; i < 6; i++) {
         std::string line;
@@ -407,9 +392,11 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string 
         if (!(LineFound)) {
             LineFound = parseInputFromList(line, NamesOfFilesOptional, OptionalFilesRead, NumOptionalInputs);
             if (!(LineFound)) {
-                std::cout << "Warning: the line " << line
-                          << " did not match any analysis file options expected by ExaCA and will be ignored"
-                          << std::endl;
+                LineFound = parseInputFromList(line, NamesOfFilesDeprecated, DeprecatedFilesRead, NumDeprecatedInputs);
+                if (!(LineFound))
+                    std::cout << "Warning: the line " << line
+                              << " did not match any analysis file options expected by ExaCA and will be ignored"
+                              << std::endl;
             }
         }
     }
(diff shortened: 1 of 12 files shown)


analysis/src/GAutils.cpp
```diff
@@ -164,9 +236,12 @@ void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax
     }
 }
 
-void ParseLogFile_Old(std::string LogFile, int &nx, int &ny, int &nz, double &deltax, int &NumberOfLayers,
-                      bool UseXYZBounds, std::vector<double> &XYZBounds) {
+void ParseLogFile_OldNoColon(std::string LogFile, int &nx, int &ny, int &nz, double &deltax, int &NumberOfLayers,
+                             bool UseXYZBounds, std::vector<double> &XYZBounds) {
 
+    std::cout << "Warning: old (non-json) form of log file detected; compatibility with analyzing these data sets will "
```
streeve, line 242: Remove (already in check format)

src/CAprint.cpp
```diff
@@ -626,7 +769,7 @@ void PrintExaCALog(int id, int np, std::string InputFile, std::string Simulation
         ExaCALog << "Nucleation density (per m^3) was: " << NMax << std::endl;
         ExaCALog << "Mean nucleation undercooling (in K, relative to liquidus temperature) was: " << dTN << std::endl;
         ExaCALog << "Standard deviation of nucleation undercooling (in K) was: " << dTsigma << std::endl;
-        ExaCALog << irf.print() << std::endl;
+        ExaCALog << irf.print_old() << std::endl;
```
streeve, line 772: `print_Old` everywhere else

analysis/src/GAutils.cpp
```diff
@@ -341,6 +416,14 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string
     LogFile = RequiredFilesRead[0];
     MicrostructureFile = RequiredFilesRead[1];
     OutputFileName = RequiredFilesRead[2];
+    std::size_t extension_pos = LogFile.find_last_of(".");
+    // Ensure that form of log file (.log or .json) matches the compilation option used to generate it
```
streeve, line 420: But there's no guarantee that you're analyzing something created from the same build of ExaCA. It seems like the issue is really that you used to add the `.log` and now you're not sure which one exists

streeve (changes requested):

analysis/src/GAutils.cpp
MattRolchigo: Yeah for running `grain_analysis` (where the input file is required, and the name of the log file including the extension is required), it should just require that the log file from the input file matches the actual log file. For running `grain_analysis_amb` (no input file used), I guess it can just check for one, then check for the other

MattRolchigo: Actually, the problem with that is that the analysis file has a hardcoded filename that is either json or log - so it'll fail in the CI if it's being tested with one and the extension used is the other. So I guess I'll just have it check for both

analysis/src/runGA_AMB.cpp
```diff
@@ -31,7 +31,11 @@ int main(int argc, char *argv[]) {
     }
     else {
         BaseFileName = argv[1];
```
streeve, line 33: Better to redefine the command line arg to include the file ending?

src/runCA.cpp
```diff
@@ -521,11 +521,31 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
...
     PrintExaCALog(id, np, InputFile, SimulationType, MyYSlices, MyYOffset, irf, deltax, NMax, dTN, dTsigma, temp_paths,
                   TempFilesInSeries, HT_deltax, RemeltingYN, deltat, NumberOfLayers, LayerHeight, SubstrateFileName,
                   SubstrateGrainSpacing, UseSubstrateFile, G, R, nx, ny, nz, FractSurfaceSitesActive, PathToOutput,
                   NSpotsX, NSpotsY, SpotOffset, SpotRadius, OutputFile, InitTime, RunTime, OutTime, cycle, InitMaxTime,
                   InitMinTime, NuclMaxTime, NuclMinTime, CreateSVMinTime, CreateSVMaxTime, CaptureMaxTime,
                   CaptureMinTime, GhostMaxTime, GhostMinTime, OutMaxTime, OutMinTime, XMin, XMax, YMin, YMax, ZMin,
                   ZMax, GrainOrientationFile);
+#else
+    if (id == 0)
+        std::cout << "Warning: log file printed using non-json format - ExaCA_ENABLE_JSON=ON will be required in a "
```
streeve, line 535: "JSON will be required in a future release" - I assume we'll remove the CMake option at the same time

analysis/src/GAutils.cpp
```diff
@@ -289,6 +355,25 @@ void ReadIgnoreField(std::ifstream &InputDataStream, int nx, int ny, int nz) {
         getline(InputDataStream, line);
 }
 
+// Check if log file format is consistent with compilation option, swap form if necessary
```
streeve, line 358: "is consistent with input, swap form if necessary. FIXME: Should be removed in a future release"

streeve (commented):
Comment/warning changes. Otherwise ready to go

analysis/src/runGA_AMB.cpp
MattRolchigo: Well the command line argument is also used to find the .vtk file, so having .json or .log be required there might be a little strange. But maybe we should require an extra command line arg that specifies if the data to be read uses .json or .log format, defaulting to .log if not given

analysis/src/GAutils.cpp
```diff
@@ -289,6 +355,25 @@ void ReadIgnoreField(std::ifstream &InputDataStream, int nx, int ny, int nz) {
...
+        LogFileUpdate = LogFile;
+    else {
+        // Use the other possible version of the log file
+        std::size_t extension_pos = LogFile.find_last_of(".");
+        std::string LogFileNamePrefix = LogFile.substr(0, extension_pos);
+        std::string LogFileExt = LogFile.substr(extension_pos, std::string::npos);
+        if (LogFileExt == ".json")
+            LogFileUpdate = LogFileNamePrefix + ".log";
+        else if (LogFileExt == ".log")
+            LogFileUpdate = LogFileNamePrefix + ".json";
```
streeve, line 372: This will break if you run the old version after outputting with the new, right?

analysis/src/runGA_AMB.cpp
streeve: Forgot about that. Should be able to use the same function as the main analysis to try either one (or only txt if json isn't enabled)

analysis/src/GAutils.cpp
MattRolchigo: You mean if I have output of both types in a directory and I run it without json enabled, and it goes to read the .json file?

analysis/src/GAutils.cpp
```diff
@@ -338,37 +410,57 @@ void ParseFilenames(std::string AnalysisFile, std::string &LogFile, std::string
...
+    else if (!(OptionalFilesRead[0].empty())) {
+#ifdef ExaCA_ENABLE_JSON
+        // If json is enabled, first check if log is in json format
+        LogFile = OptionalFilesRead[0] + ".json";
+        bool Json_found = checkFileExists(LogFile, 0, false);
+        // If json is enabled and the .json file does not exist, try the .log file
+        if (!(Json_found))
+            LogFile = OptionalFilesRead[0] + ".log";
+#else
+        LogFile = OptionalFilesRead[0] + ".log";
```
streeve, line 430: Check this exists?

src/runCA.cpp
```diff
@@ -521,11 +521,29 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
...
     PrintExaCALog(id, np, InputFile, SimulationType, MyYSlices, MyYOffset, irf, deltax, NMax, dTN, dTsigma, temp_paths,
                   TempFilesInSeries, HT_deltax, RemeltingYN, deltat, NumberOfLayers, LayerHeight, SubstrateFileName,
                   SubstrateGrainSpacing, UseSubstrateFile, G, R, nx, ny, nz, FractSurfaceSitesActive, PathToOutput,
                   NSpotsX, NSpotsY, SpotOffset, SpotRadius, OutputFile, InitTime, RunTime, OutTime, cycle, InitMaxTime,
                   InitMinTime, NuclMaxTime, NuclMinTime, CreateSVMinTime, CreateSVMaxTime, CaptureMaxTime,
                   CaptureMinTime, GhostMaxTime, GhostMinTime, OutMaxTime, OutMinTime, XMin, XMax, YMin, YMax, ZMin,
                   ZMax, GrainOrientationFile);
+#else
+    if (id == 0)
+        std::cout << "Warning: JSON will be required in a future release" << std::endl;
```
streeve, line 535: Not clear this warning is coming from the log file format

streeve (commented):
Last question

analysis/src/runGA_AMB.cpp
```diff
@@ -31,7 +31,16 @@ int main(int argc, char *argv[]) {
...
+#ifdef ExaCA_ENABLE_JSON
+        // If json is enabled, first check if log is in json format
+        LogFile = BaseFileName + ".json";
+        bool Json_found = checkFileExists(LogFile, 0, false);
```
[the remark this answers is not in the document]
MattRolchigo: We do that later for both cases. Here, we're saying there's 2 possibilities: JSON is enabled, or JSON is not enabled. If JSON is enabled, we can use either a JSON or a .log file, and JSON is preferred. So we check if the JSON file exists and if it isn't, the .log file should exist. If JSON is not enabled, we don't care if there is a JSON file, but the .log file should exist. At this point, there's only 1 possibility for the file extension that should exist. We call `CheckInputFiles` and check if that exists, and throw an error if it doesn't.

analysis/src/GAutils.cpp
MattRolchigo: 7 and 8 lines after this, we check if the file is empty or not, which accomplishes the same thing

streeve (approved):

PR #170, opened 2023-02-01
New functions to modularize cross-section MTEX data printing

MattRolchigo:
* New functions for printing inverse pole figure colored cross-section data and for collecting grain orientation histogram data (previously were done as part of other subroutines in GAprint.cpp)
* Allow printing of unimodal grain statistics on a cross-section even if no pole figure or inverse pole figure colored data is to be printed

### DIFF
--- a/analysis/src/GAprint.cpp
+++ b/analysis/src/GAprint.cpp
@@ -296,7 +296,7 @@ void PrintPoleFigureData(bool *AnalysisTypes, std::string BaseFileName, int Numb
         std::string FNameM = BaseFileName + "_PFVolumeX" + std::to_string(XMin) + "-" + std::to_string(XMax) + "Y" +
                              std::to_string(YMin) + "-" + std::to_string(YMax) + "Z" + std::to_string(ZMin) + "-" +
                              std::to_string(ZMax) + ".txt";
-        WritePoleFigureDataToFile(FNameM, NumberOfOrientations, GrainEulerAngles, GOHistogram);
+        WritePoleFigure(FNameM, NumberOfOrientations, GrainEulerAngles, GOHistogram);
     }
 }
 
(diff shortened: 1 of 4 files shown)


analysis/src/GAprint.hpp
```diff
@@ -48,6 +48,10 @@ void PrintGrainAreaData(bool *AnalysisTypes, std::string BaseFileName, double de
 void PrintPoleFigureData(bool *AnalysisTypes, std::string BaseFileName, int NumberOfOrientations, int XMin, int XMax,
                          int YMin, int YMax, int ZMin, int ZMax, ViewI3D_H GrainID, ViewI3D_H LayerID,
                          ViewF_H GrainEulerAngles);
+void WriteIPFColoredCrossSectionDataToFile(std::string BaseFileName, std::string ThisCrossSectionPlane,
```
streeve, line 51: What's the difference between "WriteToFile" and "Print"?

analysis/src/GAprint.cpp
```diff
@@ -536,6 +526,8 @@ void AnalyzeCrossSection_Bimodal(std::ofstream &QoIs, std::string BaseFileName,
 }
 
 // Analysis of data for the speified cross-section(s)
+// FIXME: This subroutine should be removed in the future, as the specific analysis modes on a given volume or
```
streeve, line 529: I'm confused - it's still used in the run files, but you don't plan to in the future?

analysis/src/GAprint.hpp
MattRolchigo: I'm trying to have "WriteToFile" mean that the subroutine opens, writes, and closes a file, while "Print" is just printing data to the screen or to a file modified by multiple subroutines (though PrintPoleFigureData also calls WritePoleFigureDataToFile...)

MattRolchigo: Probably a less wordy way of doing that. "write[n]File" and "print[n]Data" where "n" is a type of data might be better. But I also modified a bunch of print routine names in the other open PR, so maybe a third one after these are merged for renaming would be good

analysis/src/GAprint.cpp
MattRolchigo: Yeah, the note is reminding me to remove it in the future to streamline things. The unimodal and bimodal analysis I'm also planning on removing/replacing with more modular functions, and just having the file of grain IDs, sizes, etc from which someone could load it into python and do what they want with it

analysis/src/GAprint.cpp
```diff
@@ -703,6 +674,55 @@ void WritePoleFigureDataToFile(std::string Filename, int NumberOfOrientations, V
...
+                XLoc = Index1;
+                YLoc = Index2;
+                ZLoc = Index3;
+            }
+            else if (Plane == "YZ") {
+                XLoc = Index3;
+                YLoc = Index1;
+                ZLoc = Index2;
+            }
+            else {
```
streeve, line 705: Check for XZ just to be sure?

analysis/src/GAprint.cpp
```diff
@@ -578,115 +570,94 @@ void PrintCrossSectionData(int NumberOfCrossSections, std::string BaseFileName,
...
+        int NucleatedGrainCells = 0;
+        int UnmeltedCells = 0;
+        int CrossSectionSize = (Index1High - Index1Low) * (Index2High - Index2Low);
+        std::vector<int> CrossSectionGrainIDs(CrossSectionSize);
+        int Counter = 0;
+        for (int Index1 = Index1Low; Index1 < Index1High; Index1++) {
+            for (int Index2 = Index2Low; Index2 < Index2High; Index2++) {
+                // How do Index1, Index2, CrossSectionOutOfPlaneLocation correspond to GrainID(Z loc, X loc, Yloc)?
+                int ZLoc, XLoc, YLoc;
+                if (Plane == "XY") {
```
streeve, line 585: This if block could be reused in a couple places

streeve (commented):
Okay definitely last comments

analysis/src/GAprint.cpp
MattRolchigo: Well since I only use it twice, and the use starting at line 585 is in a subroutine I plan on removing in the future, so probably not worth it

analysis/src/GAprint.cpp
```diff
@@ -578,115 +570,94 @@ void PrintCrossSectionData(int NumberOfCrossSections, std::string BaseFileName,
...
-            for (int i = 0; i < NumberOfGrains; i++) {
-                for (int j = 0; j < CrossSectionSize; j++) {
-                    if (UniqueGrainIDs[i] == CrossSectionGrainIDs[j])
-                        GrainAreas[i]++;
+                else if (Plane == "YZ") {
+                    XLoc = CrossSectionOutOfPlaneLocation;
+                    YLoc = Index1;
+                    ZLoc = Index2;
+                }
+                else {
```
streeve, line 595: XZ? Error if not one of these 3?

PR #175, opened 2023-02-23
Json format for analysis input files

MattRolchigo:


MattRolchigo:
Still need to update the README. Also, @streeve , running the analysis script now requires two command line inputs instead of one, theoretically should the workflow test both cases? We have the line 

`$HOME/exaca/bin/grain_analysis AnalyzeSmallDirS.txt`

In the CI, which still works without JSON, but when running with JSON it should be

`$HOME/exaca/bin/grain_analysis AnalyzeSmallDirS.json TestProblemSmallDirS`

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,350 @@
...
+#include <cmath>
+#include <fstream>
+#include <iostream>
+#ifdef ExaCA_ENABLE_JSON
+#include <nlohmann/json.hpp>
+#endif
+#include <string>
+#include <vector>
+
+#ifdef ExaCA_ENABLE_JSON
```
streeve, line 23: Seems superfluous (can just use the one above)

analysis/examples/AnalyzeAMB.json
```diff
@@ -0,0 +1,19 @@
+[
```
streeve, line 1: I would wrap this in a top level `{}` and name this list `Regions` or something in case you want to add other things later. Or maybe these inputs should separate total, sub-volumes, planes, etc.? It's not really clear to me what each of these is on first glance

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,350 @@
...
+// analyses to be performed
+struct RepresentativeRegion {
+
+    std::string regionName;
+    std::string regionType;
+    std::string regionOrientation;
+    std::vector<double> xBounds_Meters = std::vector<double>(2);
+    std::vector<double> yBounds_Meters = std::vector<double>(2);
+    std::vector<double> zBounds_Meters = std::vector<double>(2);
+    double regionSize_Meters;
```
streeve, line 34: Why both units?

analysis/examples/AnalyzeAMMT.txt
```diff
@@ -0,0 +1,29 @@
+Used in post-processing script to analyze select ExaCA outputs
```
streeve, line 1: Was this intentionally added?

analysis/src/runGA.cpp
```diff
@@ -224,11 +216,250 @@ int main(int argc, char *argv[]) {
...
+    // Check the format of the analysis input file
+    // If not json, use old analysis routine
+    bool AnalysisJsonFormat = checkInputFileFormat(AnalysisFile, 0);
+    if (!(AnalysisJsonFormat))
+        AnalysisOld(AnalysisFile);
+    else {
+#ifdef ExaCA_ENABLE_JSON
+        // Second command line argument - base file name for microstructure data and analysis output
+        if (argc < 3) {
+            throw std::runtime_error("Error: Full path to and name of analysis file and microstructure data must be "
```
streeve, line 245: "microstructure base name"?

streeve (commented):
For the CI, you can add a copy of the "Test GA" section with `if: ${{ matrix.input == "json" }}` and the new extra commandline arg version

analysis/src/GArepresentativeregion.hpp
MattRolchigo: Since the conversion from meters to cells depends on the dimensions of the region (meters, square meters, or cubic meters), and there's situations where both are needed, it made sense to store both representations instead of converting multiple times

analysis/examples/AnalyzeAMMT.txt
MattRolchigo: Nope, that's a mistake. So was AnalysisCA5.txt

.github/workflows/CI.yml
```diff
@@ -104,4 +104,9 @@ jobs:
         working-directory: analysis/examples
         run: |
           $HOME/exaca/bin/ExaCA-Kokkos ../../examples/Inp_SmallDirSolidification.${{ matrix.input }}
-          $HOME/exaca/bin/grain_analysis AnalyzeSmallDirS.txt
+      - name: Test GA JSON
+          if: ${{ matrix.input == "json" }}
```
streeve, line 108:
```suggestion
          if: ${{ matrix.input == 'json' }}
```

.github/workflows/CI.yml
```diff
@@ -104,4 +104,9 @@ jobs:
...
+          working-directory: analysis/examples
+          run: |
+             $HOME/exaca/bin/ExaCA-Kokkos ../../examples/Inp_SmallDirSolidification.${{ matrix.input }}
```
streeve, line 111: Can just be `.json`

.github/workflows/CI.yml
MattRolchigo: @streeve tried that, still getting a syntax error. The weird thing is, line 108 where it claims the error is is identical to line  104. I've never used yaml before so I have no idea what's going on here

analysis/README.md
```diff
@@ -54,3 +55,40 @@ Z Bounds (in cells): [ZMin, ZMax]
...
+
+Once the bounds of each region are identified, the analysis options can be specified within JSON arrays under `printStats`, `printPerGrainStats`, and `printLayerwiseData`. 
+* Values specified in `printStats` will print average quantities to the console as well as to a file names `[MicrostructureBaseFilename]_[RegionName]_QoIs.txt`
+* Values specified in `printPerGrainStats` will print data for each individual grain ID to a csv file of per-grain data named `[MicrostructureBaseFilename]_[RegionName]_grains.csv`
+* Values specified in `printLayerwiseData` will be printed in additional separate files.
+
+| Option                | Allowed arrays                | Details
```
streeve, line 75: Output | Compatible options | Details?

analysis/unit_test/tstRepresentativeRegion.hpp
```diff
@@ -0,0 +1,200 @@
...
+#include <nlohmann/json.hpp>
+#include <string>
+#include <vector>
+
+namespace Test {
+//---------------------------------------------------------------------------//
+// Tests of representative region structure
+//---------------------------------------------------------------------------//
+// Initialize a representative volume from json format
+void testConstructRepresentativeRegion_Volume() {
```
streeve, line 24: These are more testing the parsing (still useful), but not the analysis itself, right?

MattRolchigo: Yeah, they test the constructor of the RepresentativeRegion struct from analysis/src/GArepresentativeregion.hpp, which is why I used this as the name. Wasn't sure if a separate file just for testing the constructor and then a different one for testing the member functions made sense

analysis/examples/AnalyzeDirS.json
```diff
@@ -0,0 +1,23 @@
+{
```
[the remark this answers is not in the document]
MattRolchigo: Fixed (line 2 of the file)

.github/workflows/CI.yml
streeve: Fixed - it was some non-standard character

.github/workflows/CI.yml
```diff
@@ -104,4 +104,9 @@ jobs:
         working-directory: analysis/examples
         run: |
           $HOME/exaca/bin/ExaCA-Kokkos ../../examples/Inp_SmallDirSolidification.${{ matrix.input }}
-          $HOME/exaca/bin/grain_analysis AnalyzeSmallDirS.txt
+      - name: Test GA JSON
+        if: ${{ matrix.input == 'json' }}
```
streeve, line 108: Shouldn't this be the single case for "txt"?

.github/workflows/CI.yml
```diff
@@ -104,4 +104,9 @@ jobs:
         working-directory: analysis/examples
         run: |
           $HOME/exaca/bin/ExaCA-Kokkos ../../examples/Inp_SmallDirSolidification.${{ matrix.input }}
-          $HOME/exaca/bin/grain_analysis AnalyzeSmallDirS.txt
```
streeve, line 107: Should be the JSON case (not removed)?

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,344 @@
...
+                PrintInversePoleFigureMapYN = RegionData["printInversePoleFigureMap"];
+            }
+            else
+                PrintInversePoleFigureMapYN = false;
+        }
+        else
+            PrintInversePoleFigureMapYN = false;
+    }
+
+    void GetRegionTypeOrientation() {
```
streeve, line 138: Isn't this a `set` method?

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,344 @@
...
+        else if (regionType == "area") {
+            regionSize_Meters = regionSize_Cells * pow(deltax, 2);
+            regionSize_Microns = regionSize_Meters * pow(10, 12);
+        }
+        else if (regionType == "volume") {
+            regionSize_Meters = regionSize_Cells * pow(deltax, 3);
+            regionSize_Microns = regionSize_Meters * pow(10, 18);
+        }
+        else
+            throw std::runtime_error("Error: Invalid region type in GetRegionSize during region construction");
```
streeve, line 293: Isn't this set internally?

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,344 @@
...
+                else if (dir == 2) {
+                    zBounds_Meters[i] = Bounds_Meters[i];
+                    zBounds_Cells[i] = Bounds_Cells[i];
+                }
+            }
+        }
+    }
+
+    // Get the size of the region in cells, and in meters/microns (either 1D, 2D, or 3D units depending on region type)
+    void GetRegionSize(double deltax) {
```
streeve, line 277: Again, isn't this a `set` method?

analysis/src/runGA.cpp
```diff
@@ -224,11 +216,265 @@ int main(int argc, char *argv[]) {
...
+                "given on the command line");
+        }
+        std::string BaseFileName = argv[2];
+        std::string LogFile = BaseFileName + ".json";
+        std::string MicrostructureFile = BaseFileName + ".vtk";
+        std::cout << "Performing analysis of " << MicrostructureFile << " , using the log file " << LogFile
+                  << " and the options specified in " << AnalysisFile << std::endl;
+
+        // Initialize Kokkos
+        Kokkos::initialize();
```
streeve, line 256: Why not initialize once?

analysis/src/runGA.cpp
```diff
@@ -224,11 +216,265 @@ int main(int argc, char *argv[]) {
...
+
+                // Output file stream for quantities of interest
+                std::ofstream QoIs;
+                std::string QoIs_fname = BaseFileNameThisRegion + "_QoIs.txt";
+                QoIs.open(QoIs_fname);
+                // Header data for QoIs file (probably just pass the whole class to the subroutines in a future release
+                // once Json is required)
+                if (representativeRegion.regionType == "area")
+                    printAnalysisHeader_Area(
+                        QoIs, representativeRegion.xBounds_Cells[0], representativeRegion.xBounds_Cells[1],
```
streeve, line 323: You pass the individual components a lot - maybe just pass the arrays?

streeve (changes requested):
Nothing major. Worth considering just passing the full region object to all these methods - upside is you don't have to pass all the separate values around all the time; downside those methods are then directly tied to the regions

analysis/src/runGA.cpp
MattRolchigo: That was done so that the same subroutines could be used with and without the "new" analysis, since the old analysis input read does not create a `representativeRegion` object. I'm planning on passing the full object in the future, I'll go back and mark these spots as TODO after removing the old analysis options in version 1.2

analysis/src/GArepresentativeregion.hpp
MattRolchigo: What do you mean? I don't have any methods with "set" in the title, all the subroutines that calculate or return a specific thing all start with the word "get"

analysis/src/GArepresentativeregion.hpp
MattRolchigo: Yeah, but without it, I get a warning about `regionSize_Microns` may be uninitialized when `if` statements are false

analysis/src/GArepresentativeregion.hpp
streeve: Noting from our conversation that `set` makes more sense here from the perspective of class data

analysis/src/GAprint.cpp
```diff
@@ -837,10 +957,84 @@ void writeIPFColoredCrossSection(std::string BaseFileName, std::string CrossSect
     GrainplotIPF.close();
 }
 
+// Print data to be read by MTEX to plot the cross-section using the inverse pole figure colormap
+// Details about the plane from the representativeRegion structure
+#ifdef ExaCA_ENABLE_JSON
+void writeIPFColoredCrossSection(std::string BaseFileNameThisRegion, RepresentativeRegion representativeRegion,
```
streeve, line 963: I think it may be worth making these class methods of the region now since they already don't work with the older version

analysis/src/runGA.cpp
```diff
@@ -16,219 +16,462 @@
...
+                }
+
+                // IPF map for area print a file named "[BaseFileNameThisRegion]_IPFCrossSectionData.txt"
+                if (representativeRegion.PrintInversePoleFigureMapYN) {
+                    writeIPFColoredCrossSection(BaseFileNameThisRegion, representativeRegion, GrainID, GrainEulerAngles,
+                                                deltax, NumberOfOrientations);
+                }
+                std::cout << "Finished analysis for region " << RegionName << std::endl;
+            } // end loop over all representative regions in analysis file
+#endif
```
streeve, line 471: Need an `#else` to print out in case JSON isn't enabled?

streeve (commented):
Pretty sure this is the last round

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,344 @@
...
+            std::string onebound = boundtype + "Bound";
+            std::string twobounds = boundtype + "Bounds";
+            if ((RegionData.contains(onebound)) && (RegionData.contains(twobounds))) {
+                std::string error =
+                    "Error: region for analysis cannot have values for both " + onebound + " and " + twobounds;
+                throw std::runtime_error(error);
+            }
+            else if ((!RegionData.contains(onebound)) && (!RegionData.contains(twobounds))) {
+                Bounds_Read[0] = 0;
+                if (Units == "Meters") {
```
streeve, line 212: I think everything that takes a string, sets values, and errors if invalid should be a separate member function (here `checkUnits`)

analysis/src/GAprint.cpp
```diff
@@ -14,72 +14,102 @@
 //*****************************************************************************/
 // Write header of unique grain ID values to stats file, and print information about the representative region to the
 // console/QoIs file
-void printAnalysisHeader(std::ofstream &QoIs, const int XLow, const int XHigh, const int YLow, const int YHigh,
-                         const int ZLow, const int ZHigh, std::vector<double> XYZBounds) {
+void printAnalysisHeader_Volume_Old(std::ofstream &QoIs, const int XLow, const int XHigh, const int YLow,
```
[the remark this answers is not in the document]
MattRolchigo: All of the routines with "_Old" in them are going to be removed after the release anyways, since more relevant/better organized stats are printed via the other routines

analysis/src/GAprint.cpp
MattRolchigo: Makes sense - so once the older version is removed, does it make sense for GAprint to be added to GArepresentativeregion entirely?

analysis/src/runGA.cpp
MattRolchigo: In this case no, it's taken care of via the if block on lines 234-235

streeve: But if it's json output from a previous run that is then analyzed with a non-json exe it would do nothing, right?

analysis/src/GArepresentativeregion.hpp
MattRolchigo: There's a lot of sections of code that look similar to this, but none of them really do the exact thing (though there's. definitely some redundancy with `printMeanSize`)

streeve: Agree, but even without reuse I think it will be easier to read splitting them out

analysis/src/runGA.cpp
MattRolchigo: Oh yeah, good point, there should be an error thrown in that case

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,557 @@
...
+        double conv = convertToMicrons(deltax, regionType);
+        for (int n = 0; n < NumberOfGrains; n++) {
+            int GrainSizeCells = std::count(GrainIDVector.begin(), GrainIDVector.end(), UniqueGrainIDVector[n]);
+            // convert to either microns, square microns, or cubic microns
+            GrainSizeVector[n] = conv * GrainSizeCells;
+        }
+        return GrainSizeVector;
+    }
+
+    // TODO: Consolidate volume/area header print routines from GAprint when non-JSON option is removed
```
streeve, line 357: Not needed anymore?

MattRolchigo: No, that's reminding me that the code in `printAnalysisHeader_Volume` and the code in the very similar print area routine that's still in GAprint.cpp as it is used with both file formats can be combined

analysis/src/GAprint.cpp
```diff
@@ -14,72 +14,75 @@
...
          << "], Y = [" << YLow << "," << YHigh << "], and Z = [" << ZLow << "," << ZHigh << "] m" << std::endl;
 }
 
 // Print information about a representative area to the console/QoIs file
-void printAnalysisHeader(std::ofstream &QoIs, const int XLow_cells, const int XHigh_cells, const int YLow_cells,
-                         const int YHigh_cells, const int ZLow_cells, const int ZHigh_cells, const double XLow_microns,
-                         const double XHigh_microns, const double YLow_microns, const double YHigh_microns,
-                         const double ZLow_microns, const double ZHigh_microns, const std::string regionName,
-                         const std::string regionOrientation) {
+void printAnalysisHeader_Area(std::ofstream &QoIs, const int XLow_cells, const int XHigh_cells, const int YLow_cells,
```
streeve, line 38: Also _Old?

analysis/src/GArepresentativeregion.hpp
streeve: Now that I read this again, can't you check for valid units one time first, then do one `if` block for the 4 bounds options? If you separate the cell vs meters into yet more smaller functions you can call them in either order to simplify more

analysis/src/GAprint.cpp
MattRolchigo: Not _Old, used by both

analysis/src/GArepresentativeregion.hpp
```diff
@@ -0,0 +1,565 @@
...
+                Bounds_Cells[0] = static_cast<int>(Bounds_Read[0]);
+                Bounds_Cells[1] = static_cast<int>(Bounds_Read[1]);
+                Bounds_Meters[0] = CoordinateMinVal + deltax * Bounds_Read[0];
+                Bounds_Meters[1] = CoordinateMinVal + deltax * Bounds_Read[1];
+            }
+            else
+                throw std::runtime_error("Error: Invalid units in ConvertBounds, should be Meters or Cells");
+
+            // Store bounds for the given direction
+            for (int i = 0; i < 2; i++) {
```
streeve, line 274: <=

streeve (commented):
Made some small changes and think I found a bug?

streeve (approved):
Anything else I would suggest can wait until things are cleaned up after the release

PR #177, opened 2023-02-27
Added plotting scripts for orientation data

MattRolchigo:
Matlab scripts and colormaps for plotting of pole figure and inverse pole figure-mapped cross-section data

Co-authored-by: Adam Creuziger adam.creuziger@nist.gov


streeve:
You have to put the `Co-authored-by` in the commit message for it to work

utilities/ColorMap4.m
```diff
@@ -0,0 +1,28 @@
...
+cmap = colormap;
+setMTEXpref('defaultColorMap',cmap);
+setMTEXpref('FontSize',20);
+%% test ODF
+%{
+savepath='.';
+
+HW=20*degree;
+
+cs = crystalSymmetry('m-3m');
```
streeve, line 18: Not sure what these commented lines are for

streeve (commented):
Should add the license header to these as well. What do you think about `utilities` subfolders?

utilities/ColorMap4.m
```diff
@@ -0,0 +1,28 @@
+% Mtex color maps
+
+% Needed to produce test ODF
+%clc
+%clear
+%close all
+%% set colomap with grayscale less than 1, colors to 4 MUD
```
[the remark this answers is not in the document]
creuzige: Which license should be used?  These came out of some of my work at NIST, but the license terms are pretty permissive (https://www.nist.gov/open/copyright-fair-use-and-licensing-statements-srd-data-software-and-technical-series-publications).  I don't know how these dovetail with the license for the rest of the ExaCA.  Thoughts?

utilities/ColorMap4.m
creuzige: The commented lines from 12 to 28 allow one to generate an ODF to test the colormap without any additional code. I'm open to removing them or keeping them.  The advantage of keeping them is that it's easy to see what the colormap 'should' look like.

creuzige:
@MattRolchigo, two items to sort out:

1. I didn't see anything in the readme file about these functions.  Should something be added there?
2. Is there an example dataset to apply the functions to and test them?

Thanks!

utilities/ColorMap4.m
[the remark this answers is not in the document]
streeve: Looks like we need to put that NIST statement in each of these files no matter what, then decide on whether we add the ExaCA header. I don't see any issue there based on what you linked

utilities/ColorMap4.m
MattRolchigo: I think it makes more sense to remove them to keep the scope of this file to colormap creation only. Though if you want, we can have a separate script that loads the colormap and creates a test ODF.

MattRolchigo:
I added a new LICENSE file in the `utilities/MTEX` folder, though I wasn't exactly sure how to combine the NIST statement and the existing code license. I also added a reference to this license in the header comments of each file in this subdirectory.
@creuzige :
1. I updated the README to reference these files, the second level of the `analysis` README contains more information
2. Good idea, I added a PR to the ExaCA-Data repository to include the MTEX output from one of our example problems (https://github.com/LLNL/ExaCA-Data/pull/3)

MattRolchigo:
@creuzige Is there anyone at NIST we can ask about the license just to double check that we can put your code under the ExaCA license? ExaCA uses an MIT license and it's open source, so I feel like the NIST permissions statement should be mutually inclusive with it, but I don't really know how these things work

creuzige:
> Is there anyone at NIST we can ask about the license just to double check that we can put your code under the ExaCA license? ExaCA uses an MIT license and it's open source, so I feel like the NIST permissions statement should be mutually inclusive with it, but I don't really know how these things work

@MattRolchigo ; if you want I can sent you over to NIST legal (likely several month lead time and may get your folks legal department involved too...), but it seems like since the NIST license permits use/modification/derivative works I don't see any issue with what you've done.

PR #178, opened 2023-02-27
Updated tasmanian example generation script to use JSON

MattRolchigo:


utilities/TasmanianTest.py
```diff
@@ -55,32 +57,36 @@
...
-        "Number of layers: 56 \n",
-        "Offset between layers: 8 \n",
-        "***Output data printing options: (Y or N) which data should be printed*** \n",
-        "Print file of grain misorientation values: Y \n",
-        "Print file of all ExaCA data: Y \n",
-        "Print default RVE output: Y \n",
-        "Debug check (reduced): N \n",
-        "Debug check (extensive): N \n"]
+        OutputData = ["{ \n",
+        "   \"SimulationType\": \"RM\",\n",
```
streeve, line 63: Would be much cleaner with surrounding single quotes instead (so you don't have to escape all the quotes)

streeve (approved):
One change and it looks good

PR #183, opened 2023-04-17
Reduce size of send/recv buffers for communication

MattRolchigo:
Reduced buffer size by using an initial guess (rather than using the max possible size as done previously), and new views to track the number of cells with associated data contained in the buffer. If a rank exceeds the buffer's max capacity, a resize of the buffers is performed across all ranks

### DIFF
--- a/src/CAghostnodes.cpp
+++ b/src/CAghostnodes.cpp
@@ -8,8 +8,105 @@
 #include "CAupdate.hpp"
 #include "mpi.h"
 
+#include <algorithm>
 #include <cmath>
 #include <vector>
+
+// Set first index in send buffers to -1 (placeholder) for all cells in the buffer, and reset the counts of number of
+// cells contained in buffers to 0s
+void ResetSendBuffers(int BufSize, Buffer2D BufferNorthSend, Buffer2D BufferSouthSend, ViewI SendSizeNorth,
+                      ViewI SendSizeSouth) {
+
+    Kokkos::parallel_for(
+        "BufferReset", BufSize, KOKKOS_LAMBDA(const int &i) {
+            BufferNorthSend(i, 0) = -1.0;
+            BufferSouthSend(i, 0) = -1.0;
+        });
+    Kokkos::parallel_for(
+        "HaloCountReset", 1, KOKKOS_LAMBDA(const int) {
+            SendSizeNorth(0) = 0;
+            SendSizeSouth(0) = 0;
+        });
+}
+
+// Count the number of cells' in halo regions where the data did not fit into the send buffers, and resize the buffers
+// if necessary, returning the new buffer size
+int ResizeBuffers(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, Buffer2D &BufferNorthRecv,
+                  Buffer2D &BufferSouthRecv, ViewI SendSizeNorth, ViewI SendSizeSouth, ViewI_H SendSizeNorth_Host,
+                  ViewI_H SendSizeSouth_Host, int OldBufSize) {
+
+    int NewBufSize;
+    SendSizeNorth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeNorth);
+    SendSizeSouth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeSouth);
+    int max_count_local = max(SendSizeNorth_Host(0), SendSizeSouth_Host(0));
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Either 20 cells or 10% of previous size as additional padding (whichever is larger)
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        int padded_capacity_percent = round(0.1 * needed_buffer_capacity);
+        int padded_capacity = std::max(20, padded_capacity_percent);
+        NewBufSize = OldBufSize + needed_buffer_capacity + padded_capacity;
+        Kokkos::resize(BufferNorthSend, NewBufSize, 8);
+        Kokkos::resize(BufferSouthSend, NewBufSize, 8);
+        Kokkos::resize(BufferNorthRecv, NewBufSize, 8);
+        Kokkos::resize(BufferSouthRecv, NewBufSize, 8);
+        // Reset count variables on device to the old buffer size
+        Kokkos::parallel_for(
+            "ResetCounts", 1, KOKKOS_LAMBDA(const int &) {
+                SendSizeNorth(0) = OldBufSize;
+                SendSizeSouth(0) = OldBufSize;
+            });
+    }
+    else
+        NewBufSize = OldBufSize;
+    return NewBufSize;
+}
+
+// Refill the buffers as necessary starting from the old count size, using the data from cells marked with type
+// ActiveFailedBufferLoad
+void RefillBuffers(int nx, int nzActive, int MyYSlices, int ZBound_Low, ViewI CellType, Buffer2D BufferNorthSend,
+                   Buffer2D BufferSouthSend, ViewI SendSizeNorth, ViewI SendSizeSouth, bool AtNorthBoundary,
+                   bool AtSouthBoundary, ViewI GrainID, ViewF DOCenter, ViewF DiagonalLength, int NGrainOrientations) {
+
+    Kokkos::parallel_for(
+        "FillSendBuffersOverflow", nx, KOKKOS_LAMBDA(const int &i) {
+            for (int k = 0; k < nzActive; k++) {
+                int GlobalCellCoordinateSouth = (k + ZBound_Low) * nx * MyYSlices + i * MyYSlices + 1;
+                int GlobalCellCoordinateNorth = (k + ZBound_Low) * nx * MyYSlices + i * MyYSlices + MyYSlices - 2;
+                if (CellType(GlobalCellCoordinateSouth) == ActiveFailedBufferLoad) {
+                    int ActiveLayerCoordinateSouth = k * nx * MyYSlices + i * MyYSlices + 1;
+                    int GhostGID = GrainID(GlobalCellCoordinateSouth);
+                    float GhostDOCX = DOCenter(3 * ActiveLayerCoordinateSouth);
+                    float GhostDOCY = DOCenter(3 * ActiveLayerCoordinateSouth + 1);
+                    float GhostDOCZ = DOCenter(3 * ActiveLayerCoordinateSouth + 2);
+                    float GhostDL = DiagonalLength(ActiveLayerCoordinateSouth);
+                    // Collect data for the ghost nodes, if necessary
+                    // Data loaded into the ghost nodes is for the cell that was just captured
+                    loadghostnodes(GhostGID, GhostDOCX, GhostDOCY, GhostDOCZ, GhostDL, SendSizeNorth, SendSizeSouth,
+                                   MyYSlices, i, 1, k, AtNorthBoundary, AtSouthBoundary, BufferSouthSend,
+                                   BufferNorthSend, NGrainOrientations);
+                    CellType(GlobalCellCoordinateSouth) = Active;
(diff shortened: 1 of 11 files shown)


src/CAghostnodes.cpp
```diff
@@ -8,31 +8,124 @@
...
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Either 20 cells or 10% of previous size as additional padding (whichever is larger)
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        int padded_capacity_percent = round(0.1 * needed_buffer_capacity);
+        int padded_capacity = std::max(20, padded_capacity_percent);
+        NewBufSize = OldBufSize + needed_buffer_capacity + padded_capacity;
+        Kokkos::resize(BufferNorthSend, NewBufSize, 8);
```
streeve, line 51: You don't need the previous values do you?

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,124 @@
...
+    SendSizeNorth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeNorth);
+    SendSizeSouth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeSouth);
+    int max_count_local = max(SendSizeNorth_Host(0), SendSizeSouth_Host(0));
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Either 20 cells or 10% of previous size as additional padding (whichever is larger)
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        int padded_capacity_percent = round(0.1 * needed_buffer_capacity);
```
streeve, line 48: Would be good to make this a function arg with a default

streeve (changes requested):

src/runCA.cpp
```diff
@@ -443,12 +488,28 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
             // Initialize active cell data structures and nuclei locations for the next layer "layernumber + 1"
             if (RemeltingYN)
                 CellTypeInit_Remelt(nx, MyYSlices, LocalActiveDomainSize, CellType, CritTimeStep, id, ZBound_Low);
```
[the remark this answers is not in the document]
MattRolchigo: What do you mean? The cell type init with remelting doesn't require accessing the halo regions as no active cells are created (melting has no occurred yet, so there is no solid-liquid interface, unlike the "no remelt" case where cells with temperature data may be initialized as liquid)

src/runCA.cpp
```diff
@@ -327,33 +359,54 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
...
-                        LayerTimeTempHistory, NumberOfSolidificationEvents, RemeltingYN);
+                        BufferNorthSend, BufferSouthSend, SendSizeNorth, SendSizeSouth, ZBound_Low, nzActive, nz,
+                        SteeringVector, numSteer, numSteer_Host, AtNorthBoundary, AtSouthBoundary,
+                        SolidificationEventCounter, MeltTimeStep, LayerTimeTempHistory, NumberOfSolidificationEvents,
+                        RemeltingYN, BufSize);
+            // Count the number of cells' in halo regions where the data did not fit into the send buffers
+            // Reduce across all ranks, as the same BufSize should be maintained across all ranks
+            // If any rank overflowed its buffer size, resize all buffers to the new size plus 10% padding
+            int OldBufSize = BufSize;
+            BufSize = ResizeBuffers(BufferNorthSend, BufferSouthSend, BufferNorthRecv, BufferSouthRecv, SendSizeNorth,
```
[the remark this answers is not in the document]
MattRolchigo: What resize+optional second resize? There's only one resize here

src/CAupdate.cpp
```diff
@@ -508,12 +516,24 @@ void CellCapture(int, int np, int, int, int, int nx, int MyYSlices, InterfacialR
...
+                        // Cell activation is now finished - cell type can be changed from TemporaryUpdate to Active
+                        CellType(GlobalD3D1ConvPosition) = Active;
+                    }
+                } // End if statement for serial/parallel code
+                else {
+                    // Cell activation is now finished - cell type can be changed from TemporaryUpdate to Active
+                    CellType(GlobalD3D1ConvPosition) = Active;
                 } // End if statement for serial/parallel code
-                // Cell activation is now finished - cell type can be changed from TemporaryUpdate to Active
-                CellType(GlobalD3D1ConvPosition) = Active;
```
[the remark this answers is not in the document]
MattRolchigo: If np = 1, the cell type is still changed to active. If np > 1, the cell type only becomes active if the buffer load was successful, otherwise it is changed into a new temporary type ActiveFailedBufferLoad to mark it

src/CAghostnodes.cpp
MattRolchigo: I actually do - RefillBuffers doesn't overwrite the existing buffer data, it only fills the new buffer data with the information from the cells of type ActiveFailedBufferLoad

src/runCA.cpp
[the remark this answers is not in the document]
streeve: Sorry, resize+optional refill if needed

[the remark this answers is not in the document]
streeve: But neither are inlined?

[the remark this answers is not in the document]
MattRolchigo: But RefillBuffers calls loadghostnodes which is inlined

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,136 @@
...
+    SendSizeNorth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeNorth);
+    SendSizeSouth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeSouth);
+    int max_count_local = max(SendSizeNorth_Host(0), SendSizeSouth_Host(0));
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Add 10% of max size as additional padding, up to MaxBufSize
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        int padded_capacity = floor(0.1 * MaxBufSize);
```
streeve, line 48: 10% of max seems too big

streeve (commented):
I added comments in the old tests, but then remembered the new test - somewhere a test should start with a buffer larger than needed and then run again with one smaller than needed

unit_test/tstKokkosInitMPI.hpp
```diff
@@ -71,17 +71,19 @@ void testSubstrateInit_ConstrainedGrowth() {
     ViewF DOCenter(Kokkos::ViewAllocateWithoutInitializing("DOCenter"), 3 * LocalActiveDomainSize);
     ViewF CritDiagonalLength(Kokkos::ViewAllocateWithoutInitializing("CritDiagonalLength"), 26 * LocalActiveDomainSize);
 
-    // Buffer sizes
-    int BufSizeX = nx;
-    int BufSizeZ = nzActive;
+    // Buffer size estimate
+    int BufSize = nx * nzActive;
```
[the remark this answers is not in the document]
MattRolchigo: We test the case where the buffer isn't large enough in testResizeRefillBuffers, the init tests take care of the case where the buffers are larger than needed

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,145 @@
...
+    SendSizeNorth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeNorth);
+    SendSizeSouth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeSouth);
+    int max_count_local = max(SendSizeNorth_Host(0), SendSizeSouth_Host(0));
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Add numcells_buffer_padding (defaults to 25) cells as additional padding, up to MaxBufSize
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        NewBufSize = OldBufSize + needed_buffer_capacity + numcells_buffer_padding;
```
streeve, line 48:
```suggestion
        NewBufSize = max_count_global + numcells_buffer_padding;
```

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,145 @@
...
+}
+
+// Reset the buffer sizes to a set value (defaulting to 25, which was the initial size) preserving the existing values
+int ResetBufferCapacity(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, Buffer2D &BufferNorthRecv,
+                        Buffer2D &BufferSouthRecv, int NewBufSize) {
+    Kokkos::resize(BufferNorthSend, NewBufSize, 8);
+    Kokkos::resize(BufferSouthSend, NewBufSize, 8);
+    Kokkos::resize(BufferNorthRecv, NewBufSize, 8);
+    Kokkos::resize(BufferSouthRecv, NewBufSize, 8);
+    return NewBufSize;
```
streeve, line 74: Why return?

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,145 @@
...
+    SendSizeSouth_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), SendSizeSouth);
+    int max_count_local = max(SendSizeNorth_Host(0), SendSizeSouth_Host(0));
+    int max_count_global;
+    MPI_Allreduce(&max_count_local, &max_count_global, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
+    if (max_count_global > OldBufSize) {
+        // Increase buffer size to fit all data
+        // Add numcells_buffer_padding (defaults to 25) cells as additional padding, up to MaxBufSize
+        int needed_buffer_capacity = max_count_global - OldBufSize;
+        NewBufSize = OldBufSize + needed_buffer_capacity + numcells_buffer_padding;
+        if (NewBufSize > MaxBufSize)
```
streeve, line 49: Not sure it's worth tracking the max size - even if you exceed it (which I can't imagine) it wouldn't hurt anything

src/CAghostnodes.cpp
```diff
@@ -8,31 +8,145 @@
...
+            SendSizeNorth(0) = 0;
+            SendSizeSouth(0) = 0;
+        });
+}
+
+// Count the number of cells' in halo regions where the data did not fit into the send buffers, and resize the buffers
+// if necessary, returning the new buffer size
+int ResizeBuffers(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, Buffer2D &BufferNorthRecv,
+                  Buffer2D &BufferSouthRecv, ViewI SendSizeNorth, ViewI SendSizeSouth, ViewI_H SendSizeNorth_Host,
+                  ViewI_H SendSizeSouth_Host, int OldBufSize, int MaxBufSize, int numcells_buffer_padding) {
```
streeve, line 36: `BufferPadding`? Not sure of the change in style

src/CAghostnodes.cpp
MattRolchigo: I believe this was to be able to call this subroutine without BufSize being specified, and allowing BufSize to be reset to the default value. But this is a confusing place to do this

src/CAghostnodes.cpp
MattRolchigo: Yeah, and if we ever reach it, something probably went very wrong...

streeve (approved):
I think this is ready. The separate buffer refill when the count is exceeded makes me want to completely remove the buffer loading from cell capture, but that should be a separate PR to understand the performance implications vs better design

PR #184, opened 2023-04-18
Require two bounds to be listed for analysis region dimensions

MattRolchigo:
Avoids potential edge cases where region bounds are set incorrectly. Also fixed some syntax errors in the JSON analysis input files


streeve (approved):

streeve:
On the point about syntax errors, it may be worth adding a test that just reads each example input file

PR #185, opened 2023-04-20
Bug fix in normalization of power law interfacial response

MattRolchigo:
Unlike the quadratic and cubic interfacial responses that have all coefficients normalized by (deltat/deltax), the exponential coefficient in V = A*(undercooling)^B + C is dimensionless and should not be normalized

Fixup of #140

### DIFF
--- a/unit_test/tstInit.cpp
+++ b/unit_test/tstInit.cpp
@@ -449,13 +449,19 @@ void testInterfacialResponse() {
             BTest = 3.12;
             CTest = 0;
             FreezingRangeTest = 26.5;
-            ExpectedV = (deltat / deltax) * (ATest * pow(LocU, (deltat / deltax) * BTest) + CTest);
+            ExpectedV = (deltat / deltax) * (ATest * pow(LocU, BTest) + CTest);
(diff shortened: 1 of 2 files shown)


unit_test/tstInit.cpp
```diff
@@ -449,13 +449,16 @@ void testInterfacialResponse() {
...
             FreezingRangeTest = 26.5;
-            ExpectedV = (deltat / deltax) * (ATest * pow(LocU, (deltat / deltax) * BTest) + CTest);
+            ExpectedV = (deltat / deltax) * (ATest * pow(LocU, BTest) + CTest);
         }
         else {
             throw std::runtime_error("File not set up for testing.");
         }
         EXPECT_DOUBLE_EQ(irf.A, ATest * 2);
-        EXPECT_DOUBLE_EQ(irf.B, BTest * 2);
+        if (file_name == "SS316.json")
```
streeve, line 458: Worth a comment that this is the only input that uses this model which doesn't normalize B

streeve (approved):

PR #188, opened 2023-04-21
Add missing path to orientation files

MattRolchigo:
Only an issue when not using default installed orientation input files, or if orientation files are not present in examples/Substrate


streeve (commented):
Should be fixed in the old versions as well?

MattRolchigo:
Yeah good catch, I forgot this piece of code was redundant between the two with one inside the class and one outside the class

streeve (approved):

PR #190, opened 2023-05-03
Option to use a powder layer on layer 0

MattRolchigo:
Previously, the baseplate would span all of layer 0, and powder would only be used on successive layers. This adds the option to have the powder layer exist on all layers


src/CAinitialize.cpp
```diff
@@ -2012,8 +2027,18 @@ void SubstrateInit_FromFile(std::string SubstrateFileName, int nz, int nx, int M
...
     if (BaseplateThroughPowder)
         BaseplateSizeZ = nz; // baseplate microstructure used as entire domain's initial condition
-    else
-        BaseplateSizeZ = nzActive; // baseplate microstructure is layer 0's initial condition
+    else {
+        if (PowderFirstLayer) {
+            BaseplateSizeZ =
+                nzActive - LayerHeight; // baseplate microstructure with powder is layer 0's initial condition
+            if ((id == 0) && (BaseplateSizeZ < 1))
+                std::cout << "Warning: Substrate data from the file be used, as the powder layer extends further in "
```
streeve, line 2035: Typo?

streeve (commented):
Should this be listed in the release notes?

MattRolchigo:
This actually does fix an edge case bug in the existing baseplate init routines, so maybe that's worth mentioning in the release notes

streeve (approved):

PR #195, opened 2023-05-12
Require JSON, removed code not compatible with JSON input files

MattRolchigo:


streeve (commented):
There are some FIXME notes in the analysis about improvements once JSON is required - I assume you want to do that in a separate PR

MattRolchigo:
Yeah, there's a bunch of moving things/rewriting that I'm planning on doing in the analysis routine, essentially putting more things into the class to clean up the function inputs. But since this PR is basically just deleting old code, I felt like also adding a bunch of code would be better to do separateyl

streeve:
Sounds good - just that one very minor change then

streeve (approved):
Easy review :)

PR #197, opened 2023-05-15
Use separate format action in CI

streeve:
Use independent format action. This will 1) run faster (rather than waiting until after the build it will check and fail right away) and 2) be more clear as to why a build failed (rather than being hidden inside the main CI build)


MattRolchigo:
So what exactly is the difference between what you have now and what was there before?

streeve:
Updated the description

PR #206, opened 2023-06-08
New problem type for solidification of a single grain in an initially liquid domain

MattRolchigo:
The temperature initialization routine was repurposed to allow initialization of both the single grain problem and the directional solidification problem, as both have unidirectional thermal gradients and cooling rates in initially liquid domains. The single grain problem is also unique in the fact that it does not continue until the active layer or the simulation domain has fully solidified, but stops when the grain reaches a simulation edge - this is accounted for in a new `IntermediateOutputAndCheck` routine that is called each time step
Also includes new unit tests for the new substrate init routine, the modified temperature init routine, and the new intermediate output and check routine


MattRolchigo:
Replaced by #227

PR #212, opened 2023-06-19
Print struct to organize variables/functions

### LINKED ISSUE #104: Printing of debug data fails when remelting is included
Print debug routine (printing of view data following initialization) needs to be modified to account for the fact that the multilayer views characterizing temperature data (CritTimeStep, UndercoolingChange, LayerID) are filled on a layer by layer basis when remelting is included (without remelting, the entirety of these views are initialized before the first layer, and the appropriate data is printed).

MattRolchigo:
Consolidates data printing variables/functions inside of the PrintData object and CAprint.hpp, and allows printing of initial fields to files for the remelting type problems (Closes #104)


streeve (commented):
Conflicts to be fixed

src/CAprint.hpp
```diff
@@ -91,4 +36,640 @@ void WriteData(std::ofstream &outstream, PrintType PrintValue, bool PrintBinary,
         outstream << PrintValue << " ";
 }
 
+// Struct to hold data printing options and functions
+template <typename MemorySpace>
```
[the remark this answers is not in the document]
MattRolchigo: I think we had this argument about the nucleation data struct and I ended up doing it this way because I couldn't think of a situation where different functions would use different memory space within a class. I don't see a reason to change that here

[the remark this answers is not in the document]
streeve: My main argument against it is that it's not needed in the class - even if you always use the same memory space it's still nice to not have to specify the template arg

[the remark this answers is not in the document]
MattRolchigo: But then don't I have to specify the template argument every time I call a member function, as opposed to just doing it at the top level for the class/when I call the constructor?

streeve (approved):
Made some changes. I'm okay with this if you are

src/CAprint.hpp
```diff
@@ -317,21 +298,18 @@ struct Print {
...
+                    printViewData(id, Grainplot, nx, ny, nz, "int", "LayerID", LayerID_WholeDomain);
+                if ((id == 0) && PrintFinalMisorientation) {
+                    auto GrainUnitVector_Host =
                         Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), GrainUnitVector);
-                    view_type_int3d_host CellType_WholeDomain(
-                        Kokkos::ViewAllocateWithoutInitializing("ViewData_WholeDomain"), 0, 0,
-                        0); // not used in this call to printGrainMisorientations but is a
-                            // function input, empty view passed as dummy argument
+                    // empty view passed as dummy argument (not used in this call to printGrainMisorientation)
+                    Kokkos::View<int ***, Kokkos::HostSpace> CellType_WholeDomain(
```
MattRolchigo, line 311: Was this change just because `view_type_int3d_host` isn't used anywhere else so it wasn't worth defining the type?

src/CAprint.hpp
```diff
@@ -413,17 +380,19 @@ struct Print {
...
     // TODO: change MPI datatype to include MPI_SHORT when LayerID is stored as a consistent type
-    template <typename Print3DViewType, typename Collect1DViewTypeHost, typename Collect1DViewTypeDevice>
-    Print3DViewType collectViewData(int id, int np, int nx, int ny, int nz, int MyYSlices, MPI_Datatype msg_type,
-                                    Collect1DViewTypeDevice ViewDataThisRank_Device) {
+    template <typename Collect1DViewTypeDevice>
+    auto collectViewData(int id, int np, int nx, int ny, int nz, int MyYSlices, MPI_Datatype msg_type,
+                         Collect1DViewTypeDevice ViewDataThisRank_Device) {
+        // View (int or float extracted from 1D View) for 3D data (no initial size, only given size/filled on rank 0)
+        using value_type = typename Collect1DViewTypeDevice::value_type;
+        Kokkos::View<value_type ***, Kokkos::HostSpace> ViewData_WholeDomain(
```
MattRolchigo, line 388: That's a neat trick, I'll have to remember that this is possible

MattRolchigo: with the .label() I mean

MattRolchigo (commented):
So if I'm understanding correctly, you changed it so that `printViewData` is called on all MPI ranks instead of just rank 0, but on other ranks it just returns immediately. And the other changes were all template-related stuff that I'm not picky about as long as I'm not the one doing it. I think this all looks good to me, let me just test a few things and then I'll approve

src/CAprint.hpp
streeve: Yeah that was my thought

streeve:
Looks like I broke something in the real runs that doesn't show up in the unit tests. Not a good idea to chang the rank 0 if statements together with all the rest..

PR #215, opened 2023-06-23
CellData struct for celltype, grainid, layerid, with subviews

MattRolchigo:
- Moved cell type/substrate init/layer init subroutines into CellData struct
- Consolidated cell data init unit tests
- Used subviews for portions LayerID, GrainID, CellType, where the subviews are indexed with respect to a layer, not the whole domain

### DIFF
--- a/src/CAcelldata.hpp
+++ b/src/CAcelldata.hpp
@@ -23,20 +23,25 @@
 #include <vector>
 
 // Data that belongs to individual cells governing the current state and various ID values
+template <typename MemorySpace>
 struct CellData {
 
+    using memory_space = MemorySpace;
+    using view_type_int = Kokkos::View<int *, memory_space>;
     using view_type_int_host = Kokkos::View<int *, layout, Kokkos::HostSpace>;
+    using view_type_float = Kokkos::View<float *, memory_space>;
+
     int NextLayer_FirstEpitaxialGrainID, BottomOfCurrentLayer, TopOfCurrentLayer;
-    ViewI GrainID_AllLayers, CellType_AllLayers, LayerID_AllLayers;
+    view_type_int GrainID_AllLayers, CellType_AllLayers, LayerID_AllLayers;
 
     // Constructor for views and view bounds for current layer
     // TODO: CellType is only needed for the current layer, and LayerID is only needed for all layers/no subview is
     // needed. Leaving them as-is for now each with the "AllLayers" and "Current layer/subview slice" to minimize
     // changes to the rest of the code to accomodate. GrainID is initialized to zeros, while others are not initialized
     CellData(int LocalDomainSize, int LocalActiveDomainSize, int nx, int MyYSlices, int ZBound_Low)
-        : GrainID_AllLayers(ViewI("GrainID", LocalDomainSize))
-        , CellType_AllLayers(ViewI(Kokkos::ViewAllocateWithoutInitializing("CellType"), LocalDomainSize))
-        , LayerID_AllLayers(ViewI(Kokkos::ViewAllocateWithoutInitializing("LayerID"), LocalDomainSize)) {
+        : GrainID_AllLayers(view_type_int("GrainID", LocalDomainSize))
+        , CellType_AllLayers(view_type_int(Kokkos::ViewAllocateWithoutInitializing("CellType"), LocalDomainSize))
+        , LayerID_AllLayers(view_type_int(Kokkos::ViewAllocateWithoutInitializing("LayerID"), LocalDomainSize)) {
 
         BottomOfCurrentLayer = ZBound_Low * nx * MyYSlices;
         TopOfCurrentLayer = BottomOfCurrentLayer + LocalActiveDomainSize;
@@ -45,10 +50,10 @@ struct CellData {
     // Initializes cell types and epitaxial Grain ID values where substrate grains are active cells on the bottom
     // surface of the constrained domain. Also initialize active cell data structures associated with the substrate
     // grains
-    void substrate_init_dirsol(int id, double FractSurfaceSitesActive, int MyYSlices, int nx, int ny, int MyYOffset,
-                               NList NeighborX, NList NeighborY, NList NeighborZ, ViewF GrainUnitVector,
-                               int NGrainOrientations, ViewF DiagonalLength, ViewF DOCenter, ViewF CritDiagonalLength,
-                               double RNGSeed) {
+    void init_substrate(int id, double FractSurfaceSitesActive, int MyYSlices, int nx, int ny, int MyYOffset,
+                        NList NeighborX, NList NeighborY, NList NeighborZ, view_type_float GrainUnitVector,
+                        int NGrainOrientations, view_type_float DiagonalLength, view_type_float DOCenter,
+                        view_type_float CritDiagonalLength, double RNGSeed) {
 
         // Calls to Xdist(gen) and Y dist(gen) return random locations for grain seeds
         // Since X = 0 and X = nx-1 are the cell centers of the last cells in X, locations are evenly scattered between
@@ -64,8 +69,8 @@ struct CellData {
         // On all ranks, generate active site locations - same list on every rank
         // TODO: Generate random numbers on GPU, instead of using host view and copying over - ensure that locations are
         // in the same order every time based on the RNGSeed
-        ViewI_H ActCellX_Host(Kokkos::ViewAllocateWithoutInitializing("ActCellX_Host"), SubstrateActCells);
-        ViewI_H ActCellY_Host(Kokkos::ViewAllocateWithoutInitializing("ActCellY_Host"), SubstrateActCells);
+        view_type_int_host ActCellX_Host(Kokkos::ViewAllocateWithoutInitializing("ActCellX_Host"), SubstrateActCells);
+        view_type_int_host ActCellY_Host(Kokkos::ViewAllocateWithoutInitializing("ActCellY_Host"), SubstrateActCells);
 
         // Randomly locate substrate grain seeds for cells in the interior of this subdomain (at the k = 0 bottom
         // surface)
@@ -78,8 +83,8 @@ struct CellData {
         }
 
         // Copy views of substrate grain locations back to the device
-        ViewI ActCellX_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellX_Host);
-        ViewI ActCellY_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellY_Host);
+        view_type_int ActCellX_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellX_Host);
+        view_type_int ActCellY_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellY_Host);
 
         // Start with all cells as liquid prior to locating substrate grain seeds
         // All cells have LayerID = 0 as this is not a multilayer problem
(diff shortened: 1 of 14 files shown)


src/CAcelldata.hpp
```diff
@@ -0,0 +1,514 @@
...
+             const int ZBound_Low, const int LocalActiveDomainSize, const int LocalDomainSize, const int LayerHeight,
+             const int NGrainOrientations, const double deltax, const double FractSurfaceSitesActive,
+             const double SubstrateGrainSpacing, const double ZMin, const double RNGSeed,
+             const double PowderActiveFraction, const std::string SimulationType, const std::string SubstrateFileName,
+             const bool UseSubstrateFile, const bool BaseplateThroughPowder, const bool PowderFirstLayer,
+             NList NeighborX, NList NeighborY, NList NeighborZ, view_type_float GrainUnitVector,
+             view_type_float DiagonalLength, view_type_float DOCenter, view_type_float CritDiagonalLength,
+             view_type_int NumberOfSolidificationEvents, double *ZMaxLayer)
+        : GrainID_AllLayers(view_type_int("GrainID", LocalDomainSize))
+        , GrainID(view_type_int(Kokkos::subview(GrainID_AllLayers, std::make_pair(0, LocalActiveDomainSize))))
```
streeve, line 53: This is constructing new views from subviews of the `AllLayers` view - I'm pretty sure you just want the subviews stored or potentially to use subviews as temporaries and/or return subviews from class methods

src/CAcelldata.hpp
```diff
@@ -0,0 +1,514 @@
...
+        , CellType_AllLayers(view_type_int(Kokkos::ViewAllocateWithoutInitializing("CellType"), LocalDomainSize))
+        , CellType(view_type_int(Kokkos::subview(CellType_AllLayers, std::make_pair(0, LocalActiveDomainSize))))
+        , LayerID_AllLayers(view_type_int(Kokkos::ViewAllocateWithoutInitializing("LayerID"), LocalDomainSize))
+        , LayerID(view_type_int(Kokkos::subview(LayerID_AllLayers, std::make_pair(0, LocalActiveDomainSize)))) {
+
+        if (SimulationType == "C")
```
streeve, line 59: I think this is the right time to create separate classes for DS and non-DS problems. Probably with a shared base class

src/CAghostnodes.cpp
```diff
@@ -72,74 +72,72 @@ void ResetBufferCapacity(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, B
...
-                int GlobalCellCoordinateSouth = (k + ZBound_Low) * nx * MyYSlices + i * MyYSlices + 1;
-                int GlobalCellCoordinateNorth = (k + ZBound_Low) * nx * MyYSlices + i * MyYSlices + MyYSlices - 2;
-                if (CellType(GlobalCellCoordinateSouth) == ActiveFailedBufferLoad) {
-                    int ActiveLayerCoordinateSouth = k * nx * MyYSlices + i * MyYSlices + 1;
-                    int GhostGID = GrainID(GlobalCellCoordinateSouth);
-                    float GhostDOCX = DOCenter(3 * ActiveLayerCoordinateSouth);
-                    float GhostDOCY = DOCenter(3 * ActiveLayerCoordinateSouth + 1);
-                    float GhostDOCZ = DOCenter(3 * ActiveLayerCoordinateSouth + 2);
-                    float GhostDL = DiagonalLength(ActiveLayerCoordinateSouth);
+                int CellCoordinateSouth = k * nx * MyYSlices + i * MyYSlices + 1;
```
streeve, line 83: Can you put this indexing in the class too?

src/CAcelldata.hpp
```diff
@@ -0,0 +1,514 @@
...
+        // Copy views of substrate grain locations back to the device
+        ViewI ActCellX_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellX_Host);
+        ViewI ActCellY_Device = Kokkos::create_mirror_view_and_copy(device_memory_space(), ActCellY_Host);
+
+        // Start with all cells as liquid prior to locating substrate grain seeds
+        // All cells have LayerID = 0 as this is not a multilayer problem
+        Kokkos::deep_copy(CellType, Liquid);
+        Kokkos::deep_copy(LayerID, 0);
+        auto CellType_local = CellType;
+        auto GrainID_local = GrainID;
```
streeve, line 114: Yeah, just create the subviews here instead of storing them

src/CAcelldata.hpp
```diff
@@ -0,0 +1,514 @@
...
+
+    // Sets up views, powder layer (if necessary), and cell types for the next layer of a multilayer problem
+    //*****************************************************************************/
+    void init_next_layer(int nextlayernumber, int id, int nx, int ny, int MyYSlices, int MyYOffset, int ZBound_Low,
+                         int LayerHeight, int LocalActiveDomainSize, bool UseSubstrateFile, bool BaseplateThroughPowder,
+                         double ZMin, double *ZMaxLayer, double deltax, double RNGSeed, double PowderActiveFraction,
+                         view_type_int NumberOfSolidificationEvents) {
+
+        // Subviews for the next layer's grain id, layer id, cell type
+        int BottomOfCurrentLayer = ZBound_Low * nx * MyYSlices;
```
streeve, line 466: I would make these member variables so you can always make the right subview as needed

streeve (changes requested):
First read-through

src/CAghostnodes.cpp
streeve: You already explained this is specific to the ghosts

src/CAcelldata.hpp
MattRolchigo: I still think that it'll be easier to do this once the rest of the code is organized into structs (particularly the inputs, which is probably the first class that should be organized into base + problem-specific data)

src/CAcelldata.hpp
```diff
@@ -0,0 +1,501 @@
...
+
+    // Determine the height of the baseplate, in CA cells
+    // The baseplate always starts at the simulation bottom (Z coordinate corresponding to ZMin), regardless of whether
+    // the first layer melts the cells at the bottom or not If BaseplateThroughPowder is true, the baseplate
+    // microstructure extends through the entire simulation domain in Z If BaseplateThroughPowder is false and
+    // PowderFirstLayer is true, the baseplate extends to the Z coordinate that is LayerHeight cells short of the top of
+    // the first layer (the final LayerHeight cells of the first layer will be initialized using the powder options
+    // specified)) If BaseplateThroughPowder is false and PowderFirstLayer is false, the baseplate extends to the top of
+    // the first layer
+    int get_baseplate_size_z(int nz, double *ZMaxLayer, double ZMin, double deltax, int LayerHeight,
```
streeve, line 161: Worth considering if this belongs here if it doesn't use any member variables

src/CAghostnodes.cpp
```diff
@@ -171,6 +170,8 @@ void GhostNodes1D(int, int, int NeighborRank_North, int NeighborRank_South, int
 
     // unpack in any order
     bool unpack_complete = false;
+    ViewI CellType = cellData.getCellTypeSubview();
```
streeve, line 173: Would use `auto` once the class is templated

streeve (changes requested):

src/CAcelldata.hpp
MattRolchigo: I think it does, it's exclusively used by functions within celldata.hpp

streeve: Well I clearly read too fast then, but in that case it's worth storing some of these things in the class instead of passing them around

MattRolchigo: Yeah this is a weird function because all of the inputs would theoretically be part of a future grid class, but this function itself is only going to be used by CellData

src/CAcelldata.hpp
```diff
@@ -0,0 +1,501 @@
...
+#include <fstream>
+#include <iostream>
+#include <random>
+#include <string>
+#include <vector>
+
+// Data that belongs to individual cells governing the current state and various ID values
+struct CellData {
+
+    using view_type_int_host = Kokkos::View<int *, layout, Kokkos::HostSpace>;
```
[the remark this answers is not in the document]
MattRolchigo: But then won't I have to do `using host_view_type = typename device_view_type::HostMirror;` every time I need it for whatever kind of view?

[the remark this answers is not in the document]
streeve: You can put that here as a class typedef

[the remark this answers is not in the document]
streeve: And actually, we should replace what we have in the nucleation class with the same (it currently uses the top-level `layout`, but doesn't really need to)

streeve (commented):
I may have missed more `auto` as well. I think this is ready, but I'm a bit concerned that I may have broken real problems, like with the unmanaged View

MattRolchigo:
If you think the unmanaged view might give us problems we can roll back that commit since it isn't a related change to this PR, that part of the code was all just copied and pasted from CAinitialize.cpp

MattRolchigo:
@streeve seems like the unmanaged view was fine, but I had an unrelated bug in the intermediate misorientation output function. Should be good to go now

streeve:
I need to take one more quick look, but I think this should wait on #217 to make sure I don't make similar mistakes again here

src/runCA.cpp
```diff
@@ -237,8 +219,9 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
     double InitTime = MPI_Wtime() - StartInitTime;
     if (id == 0)
         std::cout << "Data initialized: Time spent: " << InitTime << " s" << std::endl;
-    print.printInitExaCAData(id, np, nx, ny, MyYSlices, nzActive, deltax, XMin, YMin, ZMin, GrainID, LayerID,
-                             MeltTimeStep, CritTimeStep, UndercoolingChange);
+    print.printInitExaCAData<ViewI, ViewF>(id, np, nx, ny, MyYSlices, nzActive, deltax, XMin, YMin, ZMin,
```
streeve, line 222: Can it not deduce the template parameters?

unit_test/tstKokkosUpdate.hpp
```diff
@@ -36,9 +37,12 @@ void testNucleation() {
...
     // All cells have GrainID of 0, CellType of Liquid - with the exception of the locations where the nucleation events
     // are unable to occur
-    view_int_host GrainID_Host("GrainID_Host", LocalDomainSize);
-    view_int_host CellType_Host(Kokkos::ViewAllocateWithoutInitializing("CellType_Host"), LocalDomainSize);
-    Kokkos::deep_copy(CellType_Host, Liquid);
+    CellData<TEST_MEMSPACE> cellData(LocalDomainSize, LocalActiveDomainSize, nx, MyYSlices, ZBound_Low);
+    Kokkos::deep_copy(cellData.CellType_AllLayers, Liquid);
+    auto CellType = cellData.getCellTypeSubview();
+    ViewI_H CellType_Host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), CellType);
+    ViewI GrainID = cellData.getGrainIDSubview();
```
streeve, line 44: Can't use `ViewI` becasuse it may not match `TEST_MEMSPACE` (`ViewI_H` is technically okay, but I would still suggest changing to make it a host mirror of the local View typedef). There are probably a couple more instances of this in the tests

src/runCA.cpp
MattRolchigo: Not sure, I saw the CI failure but when I saw it was something more complicated than I thought I figured I'd deal with it later

MattRolchigo: Apparently no, it couldn't deduce the types. The compiler error was: `candidate template ignored: deduced conflicting types for parameter 'ViewTypeInt' ('View<[...], CellData<Kokkos::HostSpace>::memory_space>' vs. 'View<[...], (no argument)>')
    void printFinalExaCAData(int id, int np, int nx, int ny, int nz, int MyYSlices, int NumberOfLayers,`

MattRolchigo:
@streeve I can't figure out what's going wrong here but it's something with the template types/test types not matching I think

src/CAprint.hpp
```diff
@@ -259,19 +261,20 @@ struct Print {
     // Called on rank 0, prints intermediate values of grain misorientation for all layers up to current layer, also
     // marking which cells are liquid
     template <typename ViewTypeInt, typename ViewTypeFloat>
-    void printIntermediateGrainMisorientation(int id, int np, int cycle, int nx, int ny, int MyYSlices, int nzActive,
-                                              double deltax, double XMin, double YMin, double ZMin, ViewTypeInt GrainID,
-                                              ViewTypeInt LayerID, ViewTypeInt CellType, ViewTypeFloat GrainUnitVector,
-                                              int NGrainOrientations, int layernumber, int ZBound_Low) {
+    void printIntermediateGrainMisorientation(int id, int np, int cycle, int nx, int ny, int nz, int MyYSlices,
```
streeve, line 264: This just needs the same change as the function above, I missed this one

streeve (approved):
@MattRolchigo this is finally passing

PR #217, opened 2023-07-07
Run unit tests with multiple backends

streeve:
Not enabled for all tests yet, but worth enabling for those that are rewritten with variable memory spaces

### DIFF
--- a/src/CAnucleation.hpp
+++ b/src/CAnucleation.hpp
@@ -24,9 +24,7 @@ struct Nucleation {
 
     using memory_space = MemorySpace;
     using view_type_int = Kokkos::View<int *, memory_space>;
-    using view_type_int_host = Kokkos::View<int *, layout, Kokkos::HostSpace>;
-    using view_type_float3d = Kokkos::View<float ***, memory_space>;
-    using view_type_float3d_host = Kokkos::View<float ***, layout, Kokkos::HostSpace>;
+    using view_type_int_host = typename view_type_int::HostMirror;
 
     // Four counters tracked here:
     // 1. Nuclei_WholeDomain - tracks all nuclei (regardless of whether an event would be possible based on the layer ID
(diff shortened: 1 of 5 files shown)


src/CAnucleation.hpp
```diff
@@ -75,20 +73,20 @@ struct Nucleation {
 
     // Initialize nucleation site locations, GrainID values, and time at which nucleation events will potentially occur,
     // accounting for multiple possible nucleation events in cells that melt and solidify multiple times
+    template <class... Params>
     void placeNuclei(view_type_int MaxSolidificationEvents, view_type_int NumberOfSolidificationEvents,
-                     view_type_float3d LayerTimeTempHistory, double RNGSeed, int layernumber, int nx, int ny,
-                     int nzActive, double dTN, double dTsigma, int MyYSlices, int MyYOffset, int ZBound_Low, int id,
-                     bool AtNorthBoundary, bool AtSouthBoundary) {
+                     Kokkos::View<float ***, Params...> LayerTimeTempHistory, double RNGSeed, int layernumber, int nx,
```
MattRolchigo, line 78: So if I'm understanding this correctly, this basically hardcodes the view type as float, but allows customization in some other way?

unit_test/tstKokkosMPI.hpp
```diff
@@ -285,11 +281,11 @@ void testResizeRefillBuffers() {
 
     // Allocate device views: entire domain on each rank
     // Default to wall cells (CellType(index) = 0) with GrainID of 0
-    view_type_int GrainID("GrainID", LocalDomainSize);
-    view_type_int CellType("CellType", LocalDomainSize);
+    ViewI GrainID("GrainID", LocalDomainSize);
```
MattRolchigo, line 284: Why were some of these changed back from the test int type to the CAtypes.hpp int type, while others were changed in the opposite direction?

unit_test/tstKokkosUpdate.hpp
```diff
@@ -75,15 +79,15 @@ void testNucleation() {
...
+    auto GrainID = Kokkos::create_mirror_view_and_copy(TEST_MEMSPACE(), GrainID_Host);
     nucleation.NucleiLocations = Kokkos::create_mirror_view_and_copy(TEST_MEMSPACE(), NucleiLocations_Host);
     nucleation.NucleiGrainID = Kokkos::create_mirror_view_and_copy(TEST_MEMSPACE(), NucleiGrainID_Host);
 
     // Steering Vector
-    ViewI SteeringVector(Kokkos::ViewAllocateWithoutInitializing("SteeringVector"), LocalActiveDomainSize);
+    view_int SteeringVector(Kokkos::ViewAllocateWithoutInitializing("SteeringVector"), LocalActiveDomainSize);
     // Initialize steering vector size to 0
-    ViewI numSteer("SteeringVector", 1);
+    view_int numSteer("SteeringVector", 1);
```
MattRolchigo, line 90: Like for example, this seems like the opposite change made from what you did in `testResizeRefillBuffers`

unit_test/tstKokkosMPI.hpp
streeve: This only worked because we tested with one backend at a time (consistent with the default backend). The ghost loading isn't templated yet, so it breaks if you use a non-default backend

unit_test/tstKokkosUpdate.hpp
streeve: Right, the nucleation is all templated now, so it can use any available backend

unit_test/tstKokkosMPI.hpp
streeve: Yeah, you're right - not sure why we already changed it

MattRolchigo (approved):

PR #221, opened 2023-07-20
Update GHA checkout to v3

streeve:
CI maintenance


MattRolchigo:
Is this going to fix the fact that the CI on master fails?

MattRolchigo (approved):

streeve:
> Is this going to fix the fact that the CI on master fails?

No only the automatic push to the ExaAM mirror fails

PR #223, opened 2023-08-01
Struct for temperature data

### LINKED ISSUE #56: Consistent naming of cell coordinates

### LINKED ISSUE #204: Removal of "no remelt" problems and logic
Will require the following steps:

- [x] Remove the unneeded example files and subroutines for "no remelt" problems, adding a notice that all problems will use remelting (may require update to ExaCA-Data repository data) (also renders #95 no longer relevant)
- [X] Fix "debug"/printing of initial view data, which previously was set up only for "no remelt" problem types (also closes #104 )
- [x] Reorganize remaining views and subroutines for temperature data into a struct & Removal of redundant MeltTimeStep, CritTimeStep, and UndercoolingChange views (data is stored in LayerTimeTempHistory) #223

MattRolchigo:
* Moved the initialization and update of temperature field parameters into a new struct

* Removed `MeltTimeStep`, `CritTimeStep`, and `UndercoolingChange` views, as the data is already stored in `LayerTimeTempHistory` and is only necessary to extract when printing the data to files. Additionally, this allowed the simplification of the 3D to 1D coordinate transformation, as all views use the same coordinate systems (corresponding to the domain of the current layer of a multilayer problem). For large problems, the removal of the unneeded views and the coordinate simplification yielded a speedup of around 5%

* Reorganized unit tests, all tests now use Kokkos and the tests are organized by the corresponding hpp file of the functions tested

Closes #204; closes #56


src/CAnucleation.hpp
```diff
@@ -75,31 +76,31 @@ struct Nucleation {
     // Initialize nucleation site locations, GrainID values, and time at which nucleation events will potentially occur,
     // accounting for multiple possible nucleation events in cells that melt and solidify multiple times
     template <class... Params>
-    void placeNuclei(view_type_int MaxSolidificationEvents, view_type_int NumberOfSolidificationEvents,
-                     Kokkos::View<float ***, Params...> LayerTimeTempHistory, double RNGSeed, int layernumber, int nx,
-                     int ny, int nzActive, double dTN, double dTsigma, int MyYSlices, int MyYOffset, int, int id,
+    void placeNuclei(Temperature<device_memory_space> &temperature, double RNGSeed, int layernumber, int nx, int ny,
```
streeve, line 79: Using `device_memory_space` is the problem here - you need the class `memory_space` to be consistent. Using the top level typedefs within the new classes is going to be a problem in most cases

src/CAtemperature.hpp
```diff
@@ -0,0 +1,633 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        // Copy host view data back to device
+        MaxSolidificationEvents =
+            Kokkos::create_mirror_view_and_copy(device_memory_space(), MaxSolidificationEvents_Host);
+        LayerTimeTempHistory = Kokkos::create_mirror_view_and_copy(device_memory_space(), LayerTimeTempHistory_Host);
```
streeve, line 308: This only works because this function isn't tested with multiple spaces - should be `memory_space` for all of these

src/CAtemperature.hpp
```diff
@@ -0,0 +1,633 @@
...
+                        NumberOfSolidificationEvents_Host(coord_1d)--;
+                    }
+                }
+            }
+        }
+
+        // Copy host view data back to device
+        MaxSolidificationEvents =
+            Kokkos::create_mirror_view_and_copy(device_memory_space(), MaxSolidificationEvents_Host);
+        LayerTimeTempHistory = Kokkos::create_mirror_view_and_copy(device_memory_space(), LayerTimeTempHistory_Host);
```
streeve, line 531: Same here (could be a helper function)

unit_test/tstInitializeMPI.hpp
```diff
@@ -0,0 +1,281 @@
...
+#include "CAcelldata.hpp"
+#include "CAfunctions.hpp"
+#include "CAinitialize.hpp"
+#include "CAnucleation.hpp"
+#include "CAparsefiles.hpp"
+#include "CAtypes.hpp"
+
+#include <gtest/gtest.h>
+
+#include "mpi.h"
```
streeve, line 17: Does this test really need MPI?

unit_test/tstUpdate.hpp
```diff
@@ -0,0 +1,303 @@
...
+
+#include <gtest/gtest.h>
+
+#include <fstream>
+#include <string>
+#include <vector>
+
+namespace Test {
+//---------------------------------------------------------------------------//
+void testFillSteeringVector_Remelt() {
```
streeve, line 23: Isn't this file just `tstCellCapture` at this point?

src/CAparsefiles.hpp
```diff
@@ -57,7 +57,83 @@ ReadType ParseASCIIData(std::istringstream &ss) {
     return readValue;
 }
 std::array<double, 6> parseTemperatureCoordinateMinMax(std::string tempfile_thislayer, bool BinaryInputData);
+// Read and parse the temperature file (double precision values in a comma-separated, ASCII format with a header line -
+// or a binary string of double precision values), storing the x, y, z, tm, tl, cr values in the RawData vector. Each
+// rank only contains the points corresponding to cells within the associated Y bounds. NumberOfTemperatureDataPoints is
+// incremented on each rank as data is added to RawData
+template <typename view_type_double_host>
```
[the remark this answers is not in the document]
MattRolchigo: I think the trend I've been slowly moving towards has been using all lowercase for types, starting with a capital letter for class names or functions outside of a class, and camel case for member functions/things that will be member functions

src/CAfunctions.hpp
```diff
@@ -11,21 +11,24 @@
...
 // Assign octahedron a small initial size, and a center location
+// Note that the Y coordinate is relative to the domain origin to keep the coordinate system continuous across ranks
 template <typename ViewType>
-KOKKOS_INLINE_FUNCTION void createNewOctahedron(int D3D1ConvPosition, ViewType DiagonalLength, ViewType DOCenter,
-                                                int GlobalX, int GlobalY, int GlobalZ) {
-    DiagonalLength(D3D1ConvPosition) = 0.01;
-    DOCenter(3 * D3D1ConvPosition) = GlobalX + 0.5;
-    DOCenter(3 * D3D1ConvPosition + 1) = GlobalY + 0.5;
-    DOCenter(3 * D3D1ConvPosition + 2) = GlobalZ + 0.5;
+KOKKOS_INLINE_FUNCTION void createNewOctahedron(const int coord_1d, ViewType DiagonalLength, ViewType DOCenter,
```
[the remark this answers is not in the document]
MattRolchigo: Would `index` be better?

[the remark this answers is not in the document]
streeve: I like it better

MattRolchigo:
Yes this closes #56 as well, since that was previously blocked by the way the temperature views were set up

src/CAtemperature.hpp
MattRolchigo: Those copies won't be needed anymore once all the initialize routines are moved to the device (right now only the one for directional solidification is on the device)

unit_test/tstInitializeMPI.hpp
MattRolchigo: Nope. It used MPI before though so I just left it as-is when I reorganized the tests (after I got burned adding MPI to that readTemperatureData test just to get an unexplainable CI error despite it working locally and for all the example problems)

unit_test/tstUpdate.hpp
MattRolchigo: What do you mean? We don't have a test of cell capture, this function tests `FillSteeringVector_Remelt` like the name implies. Though my eventual goal is for all of the active cell steering vector-related operations like this function and cell capture to be members of a struct named something like `octahedra`

src/CAfunctions.hpp
```diff
@@ -111,24 +111,24 @@ KOKKOS_INLINE_FUNCTION int getGrainID(int NGrainOrientations, int MyGrainOrienta
 // Get the 1D cell coordinate from the x, y, and z cell positions
 KOKKOS_INLINE_FUNCTION int get1Dcoord(const int coord_x, const int coord_y_local, const int coord_z, const int nx,
```
streeve, line 112: Now it's half coord and half index..

unit_test/tstUpdate.hpp
streeve: I guess I thought of all of these things directly facilitating cell capture, but most any name other than "update" will be more informative

src/CAfunctions.hpp
MattRolchigo: That's what I get for doing a find and replace in every file...

MattRolchigo: I like coord_x, coord_y, and coord_z for the actual x,y,z positions of the cells though, those aren't actually used to index the 1D arrays the way index is

unit_test/tstUpdate.hpp
streeve: Fair enough, just commenting on naming since you opened the door

MattRolchigo: Yeah and there's a bunch of stuff that still has to be renamed/names of functions and variables that have to be changed to match the syntax I've started to use. I stopped myself once I realized how big this PR was getting

src/CAfunctions.hpp
MattRolchigo: But they aren't used to actual index anything unlike `index` which is a little confusing. I get what you mean, since coordinate might imply that the values are in microns or meters or something - but every position and index in the code is in cell units so I kind of thought of them as interchangeable, with the only difference being that all of the views are 1D and `index` is used to access elements, and `coord` is a 3D location

streeve: Sounds good to me

streeve (approved):
Looks good. One very minor question

PR #225, opened 2023-08-04
Use of short type for LayerID

MattRolchigo:
* Consistent use of `short` for LayerID
* Modification of analysis field parsing routine to ensure that each parsed field is the correct type, with either `int` or `short` possible for LayerID based which commit was used to generate the vtk data (though it is now always stored as `short`)
* Template analysis field parsing routines to make more adaptable for future fields that may appear in the vtk output with various names/types


streeve (approved):
LGTM

PR #227, opened 2023-08-22
New problem type for solidification of a single grain in an initially liquid domain

MattRolchigo:
The temperature initialization routine was repurposed to allow initialization of both the single grain problem and the directional solidification problem, as both have unidirectional thermal gradients and cooling rates in initially liquid domains. The single grain problem is also unique in the fact that it does not continue until the active layer or the simulation domain has fully solidified, but stops when the grain reaches a simulation edge - this is accounted for in a new `IntermediateOutputAndCheck` routine that is called each time step
Also includes new unit tests for the new substrate init routine, the modified temperature init routine, and the new intermediate output and check routine

Replaces #206 which was out of date

### DIFF
--- a/src/CAtemperature.hpp
+++ b/src/CAtemperature.hpp
@@ -193,7 +193,7 @@ struct Temperature {
                 MaxSolidificationEvents_local(0) = 1;
                 NumberOfSolidificationEvents_local(index) = 1;
                 // All cells at init undercooling
-                UndercoolingCurrent(index) = initUndercooling;
+                UndercoolingCurrent_local(index) = initUndercooling;
             });
         if (id == 0)
             std::cout << "Undercooling field initialized to = " << initUndercooling << " K for all cells" << std::endl;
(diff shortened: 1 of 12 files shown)


unit_test/tstTemperature.hpp
```diff
@@ -173,5 +274,10 @@ TEST(TEST_CATEGORY, temperature) {
         testReadTemperatureData(NumberOfLayers_vals[test_count], LayerwiseTempRead_vals[test_count],
                                 TestBinaryInputRead_vals[test_count]);
     }
+    // Test for directional and single grain problems, and with and without a thermal gradient for the single grain
+    // problem
+    testInit_UnidirectionalGradient("C", 1000000);
+    //    testInit_UnidirectionalGradient("SingleGrain", 0);
```
streeve, line 280: Why are these commented?

src/CAtemperature.hpp
```diff
@@ -169,29 +169,79 @@ struct Temperature {
...
+                LayerTimeTempHistory_local(index, 0, 0) = -1;
+                // Cells reach liquidus at a time dependent on their Z coordinate
+                LayerTimeTempHistory_local(index, 0, 1) = -1;
+                // Cells cool at a constant rate
+                LayerTimeTempHistory_local(index, 0, 2) = R * deltat;
+                // All cells solidify once
+                MaxSolidificationEvents_local(0) = 1;
+                NumberOfSolidificationEvents_local(index) = 1;
+                // All cells at init undercooling
+                UndercoolingCurrent(index) = initUndercooling;
```
streeve, line 196: Needs to be “local”

unit_test/tstCellData.hpp
```diff
@@ -24,6 +24,77 @@ namespace Test {
...
+
+    // Intialize neighbor list structures (NeighborX, NeighborY, NeighborZ)
+    NList NeighborX, NeighborY, NeighborZ;
+    NeighborListInit(NeighborX, NeighborY, NeighborZ);
+
+    // Cell data struct
+    CellData<memory_space> cellData(DomainSize, DomainSize, nx, ny_local, 0);
+
+    // Orientation data - init to dummy values
+    ViewF GrainUnitVector(Kokkos::ViewAllocateWithoutInitializing("GrainUnitVector"), 9);
```
streeve, line 64: Don’t these need to be local view typedefs?

streeve (changes requested):

unit_test/tstTemperature.hpp
MattRolchigo: Probably because I meant to mark this as draft but forgot...

src/CAupdate.cpp
```diff
@@ -627,3 +627,71 @@ void IntermediateOutputAndCheck(int id, int np, int &cycle, int ny_local, int Do
...
+            if (EdgesReached_Host(edgedim, edgename)) {
+                std::cout << EdgeNames[edgename] << " edge of domain in the " << EdgeDims[edgedim]
+                          << " direction was reached on rank " << id << " and cycle " << cycle
+                          << "; simulation is complete" << std::endl;
+                XSwitchLocal = 1;
+            }
+        }
+    }
+    // Simulation ends if a global domain boundary was reached on any rank
+    MPI_Allreduce(&XSwitchLocal, &XSwitch, 1, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
```
streeve, line 696: Just a note that `XSwitch` is not the most clear variable name

streeve (approved):

src/CAupdate.cpp
streeve: I don't know, something about the simulation ending. But definitely within a stand-alone variable rename PR

PR #228, opened 2023-08-24
Added examples/unit test for SingleGrain problem type

MattRolchigo:
@streeve on my laptop this runs in about a minute on one core, or 22 seconds on four cores; how fast do you think a "small equiaxed grain" test problem should be so that we can incorporate this problem type into the CI tests/unit tests?

### DIFF
--- a/examples/Inp_EquiaxedGrain.json
+++ b/examples/Inp_EquiaxedGrain.json
@@ -1,0 +1,32 @@
+{
+   "SimulationType": "SingleGrain",
+   "MaterialFileName": "Inconel625.json",
+   "GrainOrientationFile": "GrainOrientationVectors.csv",
+   "Domain": {
+       "CellSize": 1,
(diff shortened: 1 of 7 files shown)


streeve:
If there are values to check, I'm all for adding it as a unit test. We can always add a way to disable it for quick runs if it becomes an issue

MattRolchigo:
Yeah the value to check would be the time step where the grain hits the wall and the main loop is exited - (on 1 GPU, it basically amounts to a unit test of FillSteeringVector/CellCapture since there's no nucleation and no halo exchange or temperature data). It would be a long unit test though, the existing ones are all done in 5 seconds each max I think

streeve:
I think it's worth having. Can it be a smaller system?

MattRolchigo:
@streeve I updated this with the new example + test

streeve (approved):

Issue #229, opened 2023-08-25
Grain initialization for complex parts


jwwtc:
@MattRolchigo Hello. For a part similar to the images, or even a more complex part, does it make sense to initialize substrate grains where there is a lack of thermal data? Wouldn't it be better to skip the initialization above a certain height? Thanks in advance.

<img src="https://github.com/LLNL/ExaCA/assets/69092842/86d1f6a8-9933-458f-b698-c21fbd860b94" width="45%" align="left" alt="TwoBlock_design">
<img src="https://github.com/LLNL/ExaCA/assets/69092842/595b18c6-d7c6-4bae-b714-425c79fa6626" width="45%" align="right" alt="TwoBlock">

MattRolchigo:
@jwwtc , this is a good question, there are two ways of handling this region in ExaCA:

The first is extending the substrate through the layers, so that the "baseplate" microstructure is continuous. This has the effect of allowing new grains from the edges of the part to advance inward, and allows a single variable (substrate grain size) to control the size of these new grains in both the "layer 0" boundary condition and the "layer n" boundary condition. This can be done by setting `inputdata["Substrate"]["ExtendSubstrateThroughPower"]` to `true` in the input file. This has the downside of making substrate generation take longer, and I later came up with a better way of handling the edges of the thermal data for each layer.

The second (and in my opinion, better) way of handling this is to set the variable`["Substrate"]["PowderDensity"]` in the input file - this sets a special kind of nucleation site at these edges, where new grains nucleate once the corresponding cell is undercooled. Similar to the nucleation density, this is given as a value normalized by 10^12 /m^3 (i.e., a value of 1 will set the density of these powder surface sites to 10^12). If, for a given problem, you don't believe that new grain nucleation/growth inward from powder particles for powder bed fusion is significant at these edges, you can set this to zero.

jwwtc:
Surprisingly, I'm observing extremely similar results (as seen in the cross-sectional images) when setting the `["Substrate"]["PowderDensity"]` parameter to either `0` or `1000` in the input file. Possibly, the difference would be apparent in lower mesh/cell sizes? I could attach the temperature file as well, if the issue is not apparent from these.

### Cross-sections

#### `["PowderDensity"]=0`

![TwoBlock_0](https://github.com/LLNL/ExaCA/assets/69092842/6a864c91-e383-421e-b9fe-e042cb2340ab)

#### `["PowderDensity"]=1000`

[...]

PR #234, opened 2023-09-11
Use input baseplate top location to set powder layer start

MattRolchigo:
The current code fails to correctly initialize a powder layer in the case where `LayerHeight` = 0 for simulations using externally generated temperature data (where the layer offset is built into the temperature data itself). The new code uses a new input, `BaseplateTopZ`, to explicitly state where the baseplate ends and the powder layer(s) begin. If the new input is not given, a value of Z = 0 µm is assumed (which has been used for the example problems and most other problems to date). By specifying a baseplate top value, the different between the top of the temperature data for layer `n` and layer `n+1` can be correctly used to place the powder for layer `n+1`

Addresses #229

### DIFF
--- a/src/CAcelldata.hpp
+++ b/src/CAcelldata.hpp
@@ -143,24 +143,32 @@ struct CellData {
             std::cout << "Number of substrate active cells across all ranks: " << SubstrateActCells << std::endl;
     }
 
-    void init_substrate(std::string SubstrateFileName, bool UseSubstrateFile, bool BaseplateThroughPowder,
-                        bool PowderFirstLayer, int nx, int ny, int nz, int LayerHeight, int DomainSize,
-                        double *ZMaxLayer, double ZMin, double deltax, int ny_local, int y_offset, int z_layer_bottom,
-                        int id, double RNGSeed, double SubstrateGrainSpacing, double PowderActiveFraction,
-                        view_type_int NumberOfSolidificationEvents) {
+    void init_substrate(std::string SubstrateFileName, bool UseSubstrateFile, bool BaseplateThroughPowder, int nx,
+                        int ny, int nz, int DomainSize, double *ZMaxLayer, double ZMin, double deltax, int ny_local,
+                        int y_offset, int z_layer_bottom, int id, double RNGSeed, double SubstrateGrainSpacing,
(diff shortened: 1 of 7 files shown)


src/CAinitialize.cpp
```diff
@@ -201,16 +201,18 @@ void InputReadFromFile(int id, std::string InputFile, std::string &SimulationTyp
         }
         else
             PowderActiveFraction = 1.0; // defaults to a unique grain at each site in the powder layers
-        // Should a powder layer be initialized at the top of the baseplate for the first layer? (Defaults to false,
-        // where the baseplate spans all of layer 0, and only starting with layer 1 is a powder layer present)
-        if (inputdata["Substrate"].contains("PowderFirstLayer"))
-            PowderFirstLayer = inputdata["Substrate"]["PowderFirstLayer"];
+        if ((inputdata["Substrate"].contains("PowderFirstLayer")) && (id == 0))
```
streeve, line 204: Is this option missing in the readme?

src/CAcelldata.hpp
```diff
@@ -170,33 +178,43 @@ struct CellData {
...
     }
 
     // Initializes Grain ID values where the substrate comes from a file
     void init_baseplate_grainid(std::string SubstrateFileName, int nz, int nx, int ny_local, int y_offset, int id,
-                                double ZMin, double *ZMaxLayer, bool BaseplateThroughPowder, bool PowderFirstLayer,
-                                int LayerHeight, double deltax) {
+                                int BaseplateSizeZ) {
 
+        if (id == 0)
+            std::cout << "Warning: Reading substrate data from a file will require a vtk file of GrainID values in a "
```
streeve, line 215: Can you add this to the changelog issue? And should this be noted in the readme somewhere?

streeve (approved):

src/CAcelldata.hpp
MattRolchigo: Well, I won't remove the option in 1.3 since we didn't have this warning back when we released 1.2. I figure when I add the ability to read prior vtk data to initialize a simulation, I'll note that in the README

streeve: Right, but we track both deprecation and removal of previously deprecated

PR #236, opened 2023-09-27
Fixed edge case where int overflow occurs calculating the number of potential nuclei in a layer

### LINKED ISSUE #235: Integer overflow for very large problem sizes
As the current code collects data from all cells on rank 0 to print to files, if the total number of cells across all ranks is larger than INT_MAX, this will fail. Will require updating `int` variables to `long int` in any subroutines where the total number of cells across all ranks is calculated or used

MattRolchigo:
One part of #235, though other portions of the code also will require similar fixes to fully fix

### DIFF
--- a/src/CAnucleation.hpp
+++ b/src/CAnucleation.hpp
@@ -100,13 +100,25 @@ struct Nucleation {
         std::normal_distribution<double> Gdistribution(dTN, dTsigma);
 
         // Max number of nucleated grains in this layer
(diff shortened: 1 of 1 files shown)


streeve:
This isn't enough to run on Frontier with 100 nodes. I need to bisect my local changes to find what we're missing here

streeve (approved):
Worked on 1000 nodes

PR #237, opened 2023-09-28
fixup: add missing parallel labels

streeve:


MattRolchigo (approved):

streeve:
@MattRolchigo do you want to fight with getting the HIP build to run or just merge this?

MattRolchigo:
Why is the HIP build failing anyways, all you did was change the label...

streeve:
If you click on details it shows that the initial stage quit because it ran out of space. We need to figure out how to fix it, but right now it fails randomly

PR #240, opened 2023-10-13
Deprecation notice for spot melt array problem type (S)

MattRolchigo:


streeve (approved):
Looks good - can you add to the release notes draft?

PR #241, opened 2023-10-13
Ability to read temperature data with additional columns

MattRolchigo:
This will allow ExaCA to read data created using finch with the extra Gx, Gy, Gz columns (or any other arbitrary extra columns), as long as the first 6 columns are the x, y, z, tm, tl, cr data required for simulation type R

### DIFF
--- a/src/CAparsefiles.cpp
+++ b/src/CAparsefiles.cpp
@@ -91,7 +91,7 @@ void splitString(std::string line, std::vector<std::string> &parsed_line, std::s
 std::size_t checkForHeaderValues(std::string header_line) {
 
     // Header values from file - number of commas plus one is the size of the header
-    size_t header_size = std::count(header_line.begin(), header_line.end(), ',') + 1;
+    std::size_t header_size = std::count(header_line.begin(), header_line.end(), ',') + 1;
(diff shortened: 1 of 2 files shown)


src/CAparsefiles.cpp
```diff
@@ -85,21 +85,24 @@ void splitString(std::string line, std::vector<std::string> &parsed_line, int ex
...
 }
 
-// Check to make sure that all expected column names appear in the header for this temperature file
-void checkForHeaderValues(std::string header_line) {
+// Check to make sure that the 6 expected column names appear in the correct order in the header for this temperature
+// file Return the number of columns present - ignore any columns after the 6 of interest
+int checkForHeaderValues(std::string header_line) {
 
-    // Header values from file
-    std::size_t header_size = 6;
```
streeve, line 92: Why remove the `std`?

streeve (changes requested):

src/CAparsefiles.cpp
MattRolchigo: Probably just left it off by accident, I added it back on

MattRolchigo: CAparsefiles.cpp line 94?

PR #242, opened 2023-10-16
Add CUDA build CI 

streeve:


streeve:
@MattRolchigo finally added a cuda build (to be clear it doesn't run the tests, but better than the nothing we've had so far)

MattRolchigo (approved):

PR #243, opened 2023-10-17
Fix initialization of buffer capacity during resize

MattRolchigo:
Previously, the new positions in `BufferNorthSend` and `BufferSouthSend` after resizing were left uninitialized - these should be initialized to -1 so that positions that do not hold cell data are correctly interpreted as "empty" in `GhostNodes1D`

### DIFF
--- a/src/CAghostnodes.cpp
+++ b/src/CAghostnodes.cpp
@@ -55,6 +55,14 @@ int ResizeBuffers(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, Buffer2D
                 SendSizeNorth(0) = OldBufSize;
                 SendSizeSouth(0) = OldBufSize;
             });
+        // Set -1 values for the new (currently empty) positions in the resized buffer
+        Kokkos::parallel_for(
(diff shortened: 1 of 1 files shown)


src/CAghostnodes.cpp
```diff
@@ -55,6 +55,14 @@ int ResizeBuffers(Buffer2D &BufferNorthSend, Buffer2D &BufferSouthSend, Buffer2D
                 SendSizeNorth(0) = OldBufSize;
                 SendSizeSouth(0) = OldBufSize;
             });
+        // Set -1 values for the new (currently empty) positions in the resized buffer
+        Kokkos::parallel_for(
+            "InitNewBufCapacity", Kokkos::RangePolicy<>(OldBufSize, NewBufSize), KOKKOS_LAMBDA(const int &BufPosition) {
```
streeve, line 60: Let's use the default execution space (rather than empty template input) from the View memory space - that will make it an easier transition when this is converted to a class

MattRolchigo: What do you mean? We have an empty template input for every other parallel_for in the code, including the ones in struct-containing .hpp files

streeve: Fair - we shouldn't, but it is a separate issue

streeve (approved):

PR #244, opened 2023-10-18
Use default exec_space from the memory_space, not overall default

streeve:
Otherwise `Kokkos::DefaultExecutionSpace` is always used. This guarantees that the execution space is compatible with the memory space

### DIFF
--- a/src/CAcelldata.hpp
+++ b/src/CAcelldata.hpp
@@ -33,6 +33,9 @@ struct CellData {
     using view_type_short = Kokkos::View<short *, memory_space>;
     using view_type_float = Kokkos::View<float *, memory_space>;
 
+    // Using the default exec space for this memory space.
+    using execution_space = typename memory_space::execution_space;
+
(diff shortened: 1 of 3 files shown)


src/CAcelldata.hpp
```diff
@@ -59,10 +59,15 @@ struct CellData {
         int grainLocationY = floorf(static_cast<float>(ny) / 2.0);
         int grainLocationZ = floorf(static_cast<float>(nz) / 2.0);
 
+        // Local copies for lambda capture.
         auto CellType_AllLayers_local = CellType_AllLayers;
         auto GrainID_AllLayers_local = GrainID_AllLayers;
+
+        // Using the default exec space for this memory space.
+        using exec_space = typename memory_space::execution_space;
```
MattRolchigo, line 67: Probably not worth having this line every time/only having this at the top level of the structs since we don't do anything with execution spaces

streeve: Updated

MattRolchigo (approved):

PR #245, opened 2023-10-18
Fix CMake to make ExaCA a usable downstream library

streeve:
- JSON variable was removed
- Target was not exported
- Some headers missing
- Moved files with main functions to `bin/`


MattRolchigo (approved):

streeve:
Looking into the analysis library now

PR #247, opened 2023-10-19
Additional customization of problem type C (directional solidification)

MattRolchigo:
1) As an alternative to assigning a fraction of the cells at the bottom domain surface as epitaxial grains, lists of X coordinates, Y coordinates, and GrainID values can be used to initialize this problem type.
2) The magnitude of the initial undercooling of the bottom surface can be set at values other than the liquidus
3) The liquid cells between the epitaxial grains along the bottom surface can optionally be assigned the GrainID of the nearest initialized grain, yielding an initial solidification front that consists of all cells along the bottom surface


examples/Inp_TwoGrainDirSolidification.json
```diff
@@ -0,0 +1,41 @@
...
+      "MeanUndercooling": 5,
+      "StDev": 0.5
+   },
+   "TemperatureData": {
+      "InitUndercooling": 10,
+      "G": 500000,
+      "R": 300000
+   },
+   "Substrate": {
+      "GrainLocationsX": [100, 100],
```
streeve, line 24: Any reason this can't be a list of lists? `"GrainLocations": [ [100, 50, 0], [100, 150, 0] ]`

src/CAinputs.hpp
```diff
@@ -240,14 +243,51 @@ struct Inputs {
...
+            // Must contain FractionSurfaceSitesActive OR all of GrainLocationsX/GrainLocationsY/GrainIDs, not both
+            bool contains_FractionSurfaceSitesActive = inputdata["Substrate"].contains("FractionSurfaceSitesActive");
+            bool contains_GrainLocationsX = inputdata["Substrate"].contains("GrainLocationsX");
+            bool contains_GrainLocationsY = inputdata["Substrate"].contains("GrainLocationsY");
+            bool contains_GrainIDs = inputdata["Substrate"].contains("GrainIDs");
+            bool contains_CustomGrainInitInfo =
+                contains_GrainLocationsX && contains_GrainLocationsY && contains_GrainIDs;
+            if ((contains_FractionSurfaceSitesActive) &&
+                (contains_GrainLocationsX || contains_GrainLocationsY || contains_GrainIDs))
+                throw std::runtime_error(
```
streeve, line 265: More permissive option: warn that you're going to ignore that value

examples/Inp_TwoGrainDirSolidification.json
MattRolchigo: Mostly just because I thought it was cleaner to specify lists of X and Y locations, feel like lists of lists have a higher probability of having a comma or brace left off accidentally

src/CAinputs.hpp
streeve: I just meant ignoring the surface sites or vice versa rather than errors. In your example I would error (but combining x and y reduces that chance..)

examples/Inp_TwoGrainDirSolidification.json
streeve: Disagree, but up to you

MattRolchigo: What if we make it a list of lists, but also throw an error if the lists aren't all the same length (and also the same length as the number of grain IDs)?

src/CAinputs.hpp
MattRolchigo: Fair enough, I can let the lists of specific grain data override the surface sites input and throw a warning that its ignoring FractSurfaceSitesActive

examples/Inp_TwoGrainDirSolidification.json
streeve: Agree that it has to be same length as the IDs, but I meant a list of pairs to force x,y to always match

MattRolchigo: I changed my mind again on this, I don't think I like the fact that grain locations would be a list of lists but then GrainIDs is its own list. I'd rather just have 3 separate inputs, each consisting of a list of the same size

streeve: Ok

examples/Inp_TwoGrainDirSolidification.json
```diff
@@ -0,0 +1,41 @@
...
+      "GrainLocationsY": [50, 150],
+      "GrainIDs": [25, 9936],
+      "FillBottomSurface": false
```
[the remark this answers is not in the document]
MattRolchigo: It could be true or false but for the purpose of being an example problem, I think it's more intuitive to watch the two octahedra at the specified locations grow and impinge rather than fill the entire bottom surface with active sites

streeve (approved):

PR #253, opened 2023-11-16
Grid and interface structs

MattRolchigo:


src/CAinterface.hpp
```diff
@@ -0,0 +1,510 @@
...
+                            printf("Error: Send/recv buffer resize failed to include all necessary data, predicted "
+                                   "results at MPI processor boundaries may be inaccurate\n");
+                    }
+                }
+            });
+        Kokkos::fence();
+    }
+
+    // 1D domain decomposition: update ghost nodes with new cell data from Nucleation and CellCapture routines
+    void halo_update(int, int, Grid &grid, CellData<device_memory_space> &cellData, int NGrainOrientations,
```
streeve, line 255: Duplicate?

src/runCA.cpp
```diff
@@ -19,84 +19,35 @@ void RunProgram_Reduced(int id, int np, std::string InputFile) {
...
-    // These parameters from the inputs struct will eventually be stored in the grid struct, but for now are temporarily
-    // kept in the current scope
-    int nx = inputs.domain.nx;
-    int ny = inputs.domain.ny;
-    int nz = inputs.domain.nz;
-    int NumberOfLayers = inputs.domain.NumberOfLayers;
-    int LayerHeight = inputs.domain.LayerHeight;
-    double deltax = inputs.domain.deltax;
-    double HT_deltax = inputs.temperature.HT_deltax;
+    //  Final time step at which solidification is expected TODO: may not be needed
```
streeve, line 22: Is this needed?

src/CAcelldata.hpp
```diff
@@ -48,46 +48,41 @@ struct CellData {
...
-        BottomOfCurrentLayer = z_layer_bottom * nx * ny_local;
-        TopOfCurrentLayer = BottomOfCurrentLayer + DomainSize;
-        LayerRange = std::make_pair(BottomOfCurrentLayer, TopOfCurrentLayer);
-    }
+        , _inputs(inputs) {}
 
     // Initializes the single active cell and associated active cell data structures for the single grain at the domain
     // center
-    void init_substrate(int id, int nx, int ny, int nz, int ny_local, int y_offset, int DomainSize) {
+    void init_substrate(int id, Grid &grid) {
```
streeve, line 59: Good time to think about making as many inputs `const` as possible - would be nice to be sure that the grid isn't changed in this method

src/CAcelldata.hpp
```diff
@@ -577,11 +565,52 @@ struct CellData {
                       << TotalMeltPoolCellCount << std::endl;
     }
 
+    // Stores/returns the volume fraction of nucleated grains to the console
+    // Moved from CAfunctions.hpp
+    float calcVolFractionNucleated(int id, Grid &grid) {
```
streeve, line 570: Are we using `CamelCase` or `non_camel_case`? I think we need a separate PR to do this consistently, but may as well try to keep new changes the same

src/runCA.cpp
MattRolchigo: Needed? No. Useful to remind me that I need to reevaluate whether or not that variable has to exist and what struct it should belong to? Yes

MattRolchigo:
@streeve I can confirm that this runs correctly for the example problems on Frontier with performance consistent with what we have on master

src/CAinterface.hpp
```diff
@@ -0,0 +1,301 @@
...
+#ifndef EXACA_INTERFACE_HPP
+#define EXACA_INTERFACE_HPP
+
+#include "CAcelldata.hpp"
+#include "CAconfig.hpp"
+#include "CAfunctions.hpp"
+#include "CAinputs.hpp"
+#include "CAparsefiles.hpp"
+#include "CAtemperature.hpp"
+#include "CAtypes.hpp"
```
streeve, line 15: Do we need the types here? All the views should be internally defined

streeve (approved):
Looks good, couple small things that aren't critical

MattRolchigo:
@streeve I can't figure out why this is failing, it's literally comparing two identical strings and returning false but only when using a threaded backend. It works locally using either the string compare or the == operator

PR #254, opened 2023-11-17
Store final undercooling for all cells in multilayer domain

MattRolchigo:
Previously, only the undercooling of cells in the final layer was stored. This uses a subview so that only the undercooling of cells in the current layer of a multilayer problem is used by the update routines, but the undercooling of cells from previous layers remains stored and can be printed at the end of the simulation.

### DIFF
--- a/src/CAtemperature.hpp
+++ b/src/CAtemperature.hpp
@@ -44,7 +44,8 @@ struct Temperature {
     // A counter for the number of times each CA cell has undergone solidification so far this layer
     view_type_int SolidificationEventCounter;
     // The current undercooling of a CA cell (if superheated liquid or hasn't undergone solidification yet, equals 0)
-    view_type_float UndercoolingCurrent;
+    // Also maintained for the full multilayer domain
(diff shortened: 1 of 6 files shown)


streeve:
Should just need to rebase this

src/CAtemperature.hpp
```diff
@@ -563,6 +566,13 @@ struct Temperature {
                       << " through " << grid.nz_layer + grid.z_layer_bottom - 1 << " of the global domain" << std::endl;
     }
 
+    // Get the subview associated with the undercooling of cells in the current layer. Do not reset the undercooling of
+    // cells from the prior layer to zero as this information will be stored for a potential print (and a cell that
+    // remelts in the current layer will have its undercooling reset to 0 and recalculated)
+    void get_current_layer_undercooling(std::pair<int, int> LayerRange) {
```
streeve, line 572: Weird to have a `void` return `get` function

streeve (approved):

Issue #256, opened 2023-12-05
Download failed for JSON dependency


GuoChi-Li:
Dear Developers,

I encountered an error when I followed the instructions on the readme. Here is the code I used in the terminal.

# Building Kokkos

```
# Change this path to Kokkos source
cd ./kokkos
mkdir build
cd build
# Check the CPU architecture flag matches the hardware
cmake \
  -D CMAKE_BUILD_TYPE="Release" \
  -D CMAKE_INSTALL_PREFIX=install \
  -D Kokkos_ENABLE_OPENMP=ON \
  -D Kokkos_ARCH_INTEL_GEN=ON \
  -D Kokkos_ENABLE_CUDA=ON \
  -D Kokkos_ENABLE_CUDA_LAMBDA=ON \
  -D Kokkos_ENABLE_OPENMP=ON \
  .. ;
make install
cd ../..
```

# Building ExaCA
```
# clone the project
git clone https://github.com/LLNL/ExaCA.git

# Change this path to Kokkos installation
export KOKKOS_INSTALL_DIR=/home/flag/kokkos/build/install

# Change this path to ExaCA source
cd ./ExaCA
mkdir build
cd build
cmake \
  -D CMAKE_BUILD_TYPE="Release" \
  -D CMAKE_PREFIX_PATH=$KOKKOS_INSTALL_DIR \
  -D CMAKE_INSTALL_PREFIX=install \
  ..;
make install
cd ../..
```
# The error while executing this code
```
-- The CXX compiler identification is GNU 10.5.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Enabled Kokkos devices: OPENMP;CUDA
CMake Warning at /home/flag/kokkos/build/install/lib/cmake/Kokkos/KokkosConfigCommon.cmake:30 (MESSAGE):
  The installed Kokkos configuration does not support CXX extensions.
  Forcing -DCMAKE_CXX_EXTENSIONS=Off
Call Stack (most recent call first):
  /home/flag/kokkos/build/install/lib/cmake/Kokkos/KokkosConfig.cmake:63 (INCLUDE)
  CMakeLists.txt:7 (find_package)


[...]

PR #262, opened 2023-12-13
readme: rewrite build instructions without make

streeve:
Partially in response to #256, I find this more clear

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,19 +36,19 @@ First, if Kokkos is not already built on your system, build Kokkos:
 ```
 # Change this path to Kokkos source
 cd ./kokkos
-mkdir build
-cd build
-# Check the CPU architecture flag matches the hardware
+# Configure Kokkos
 cmake \
+  -B build \
   -D CMAKE_BUILD_TYPE="Release" \
   -D CMAKE_INSTALL_PREFIX=install \
-  -D Kokkos_ENABLE_OPENMP=ON \
-  -D Kokkos_ARCH_POWER9=ON \
-  .. ;
-make install
-cd ../..
+  -D Kokkos_ENABLE_OPENMP=ON
+# Build Kokkos
+cmake --build build
+# Install Kokkos
+cmake --install build
+cd ../
 ```
-Note that there are other host backends available. The Kokkos architecture flag must match the hardware you run on and will improve performance, if used.
+Note that there are other host backends available. The Kokkos architecture flag must match the hardware you run on and will improve performance, if used (e.g. -DKokkos_ARCH_POWER9=ON).
 
 Then build ExaCA, including the path to the Kokkos build:
 ```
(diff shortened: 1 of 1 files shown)


MattRolchigo:
@streeve When I compile Kokkos using these instructions, I don't get an `install` directory inside of `build`

streeve:
> When I compile Kokkos using these instructions, I don't get an `install` directory inside of `build`

My bad, should be fixed

README.md
```diff
@@ -34,38 +34,44 @@ ExaCA has been tested with Serial, OpenMP, Pthreads, CUDA, and HIP backends.
 
 First, if Kokkos is not already built on your system, build Kokkos:
 ```
+# Change this path to desired Kokkos installation location
+export KOKKOS_INSTALL_DIR=`pwd`/install/kokkos
```
MattRolchigo, line 38: Shouldn't this be `pwd`/kokkos/install?

README.md
```diff
@@ -34,38 +34,44 @@ ExaCA has been tested with Serial, OpenMP, Pthreads, CUDA, and HIP backends.
...
   -D CMAKE_PREFIX_PATH=$KOKKOS_INSTALL_DIR \
-  -D CMAKE_INSTALL_PREFIX=install \
-  ..;
-make install
-cd ../..
+  -D CMAKE_INSTALL_PREFIX=$EXACA_INSTALL_DIR
+# Build ExaCA
+cmake --build build
+# Install ExaCA
+cmake --install build
```
MattRolchigo, line 73: Same issue here as with kokkos, no install directory is created

MattRolchigo:
So your update puts the `install` directory inside of `kokkos` and `ExaCA` themselves, instead of inside the `build` folders?

README.md
```diff
@@ -34,38 +34,44 @@ ExaCA has been tested with Serial, OpenMP, Pthreads, CUDA, and HIP backends.
...
+Note that there are other host backends available. The Kokkos architecture flag must match the hardware you run on and will improve performance, if used (e.g. -DKokkos_ARCH_POWER9=ON).
 
 Then build ExaCA, including the path to the Kokkos build:
 ```
-# Change this path to desired Kokkos installation location
-export KOKKOS_INSTALL_DIR=./kokkos/build/install
+# Change this path to desired ExaCA installation location
+export EXACA_INSTALL_DIR=`pwd`/install/exaca
+# Change this path to Kokkos installation location
+export KOKKOS_INSTALL_DIR=`pwd`/install/kokkos
```
MattRolchigo, line 61: When I build kokkos, I get a separate `build` and `install` directory, but there's no `kokkos` subdirectory inside of `install`

MattRolchigo: Okay I got it to compile, though shouldn't we have something about the gtest_install being included in the cmake since we talk about the cmake flag for enabling testing?

MattRolchigo: Will this affect the tests/example problems that rely on files in the install directories?

streeve: 1. We can put more detail about gtest, but I don't think it's critical
2. No, the point is that you can install anywhere - the build knows where it installed to (and I probably made it confusing by just putting inside build/ previously)

streeve:
> So your update puts the `install` directory inside of `kokkos` and `ExaCA` themselves, instead of inside the `build` folders?

Well inside a top-level `install/app/`, but yes, separate from `build`

MattRolchigo:
For workflows where the path to the ExaCA executable is hardcoded, this will break things though. Could we have a copy of the install directory/executable inside of build with a warning that after the v1.3 release, the `build/install` path will no longer be valid?

streeve:
> For workflows where the path to the ExaCA executable is hardcoded, this will break things though. Could we have a copy of the install directory/executable inside of build with a warning that after the v1.3 release, the `build/install` path will no longer be valid?

To be clear - this doesn't change anything at all about how ExaCA builds. This is a change in the **readme only** for suggested builds. The whole point of out of source builds is that it can be done anywhere (and no matter what you can install anywhere)

A lot of the issues I see with the current instructions come down to what/when what folder you're in, which (I think) is exacerbated by the `make` backend usage.

MattRolchigo (approved):

PR #264, opened 2023-12-14
Fixup view resize for layer time-temperature history

MattRolchigo:
Use layer-specific number of melt/solidification events when resizing this view - otherwise this may seg fault if layer "n" has more melt/solidification events than layer 0


streeve (approved):

streeve:
Add to the changelog

MattRolchigo:
This bug was actually introduced in #223 (I'm not really sure how it hasn't caused issues until now...), which was after the v1.2 release. I don't think it's worth noting in the changelog that we temporarily created and then fixing a bug that didn't exist in v1.2 to begin with

streeve:
Good to know

Issue #260, opened 2023-12-13
Intermittent failure of CI on testSmallEquiaxedGrain


MattRolchigo:
First observed in #253, but possibly an issue earlier than that: testSmallEquiaxedGrain may fail on the check of the number of time steps taken to finish the run of `Inp_SmallEquiaxedGrain.json` - the problem sometimes takes 1 time step longer than expected to complete

PR #265, opened 2023-12-14
Add +/-1 time step tolerance to equiaxed grain unit test

MattRolchigo:
Temporary workaround for #260


streeve (changes requested):
Let's add a FIXME comment with a link to the issue (and leave it open)

streeve (approved):

PR #269, opened 2023-12-15
Struct for grain orientation data

### LINKED ISSUE #133: Subroutine organization
Reorganize subroutines in CAinitialize and CAfunctions so finding specific functions is more intuitive

### LINKED ISSUE #226: Some unit tests use MPI unnecessarily

MattRolchigo:
Closes #133 , as the files CAinitialize.cpp/.hpp and CAfunctions.cpp/.hpp have been removed and their functions relocated to the appropriate structs. Also closes #226


MattRolchigo:
@streeve didn't this issue with the CI kokkos compilation come up before?

src/CAorientation.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+    view_type_float_host get_euler_angles() {
+        return Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), grain_bunge_euler);
+    }
+
+    view_type_float_host get_rgb_ipfz() {
+        return Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), grain_rgb_ipfz);
+    }
+
+    // Return host copies of grain euler angles and RGB values
```
streeve, line 173: Wrong place?

src/CAprint.hpp
```diff
@@ -429,11 +426,12 @@ struct Print {
                     if (GrainID_WholeDomain(k, i, j) == 0)
                         IntPrintVal = 200;
                     else {
-                        int MyOrientation = getGrainOrientation(GrainID_WholeDomain(k, i, j), NGrainOrientations);
+                        int my_orientation =
+                            get_grain_orientation(GrainID_WholeDomain(k, i, j), orientation.n_grain_orientations);
                         if (GrainID_WholeDomain(k, i, j) < 0)
-                            IntPrintVal = static_cast<short>(std::round(GrainMisorientation(MyOrientation)) + 100);
+                            IntPrintVal = static_cast<short>(std::round(grain_misorientation(my_orientation)) + 100);
```
streeve, line 432: Just a side note, but this is my main issue with using `lower_case` for both variables and functions - I can't tell on first glance if this is indexing the view or calling a free function

src/CAorientation.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+    // Get grain orientations from the specified file and the number of grain orientations and return a host view
+    // containing the data
+    view_type_float_host get_orientations_from_file(const std::string orientation_file, const int vals_per_line) {
+
+        // Read file of grain orientations
+        std::ifstream orientation_input_stream;
+        orientation_input_stream.open(orientation_file);
+
+        // Line 1 is the number of orientation values to read (if not specified already, is -1. Otherwise, check to
```
streeve, line 72: Why not just read the value? It looks like this will always have -1, then replace with the read value

src/CAorientation.hpp
```diff
@@ -0,0 +1,211 @@
...
+    using view_type_float_host = typename view_type_float::HostMirror;
+
+    // Using the default exec space for this memory space.
+    using execution_space = typename memory_space::execution_space;
+
+    // Grain orientation data views - either rotation matrices (9 vals per orientation), Euler angles (3 vals per
+    // orientation), or RGB values (3 vals per orientation) Host copy of grain_unit_vector maintained as it will be
+    // needed for all intermediate output steps (host copies of other representations only used in analysis script,
+    // obtained via member functions)
+    int n_grain_orientations = -1;
```
streeve, line 34: I don't think you need this default

src/CAorientation.hpp
```diff
@@ -0,0 +1,211 @@
...
+                if (misorientation < misorientation_angle_min) {
+                    misorientation_angle_min = misorientation;
+                }
+            }
+            grain_misorientation(n) = misorientation_angle_min;
+        }
+        return grain_misorientation;
+    }
+
+    view_type_float_host get_euler_angles() {
```
streeve, line 165: I would either rename to make it clear this is a host mirror or just return the view and make the user copy it

analysis/src/GAutils.cpp
```diff
@@ -46,55 +43,10 @@ void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax
...
-    checkFileNotEmpty(RotationFilename);
-    checkFileNotEmpty(EulerAnglesFilename);
-    checkFileNotEmpty(RGBFilename);
+    grain_unit_vector_file = logdata["GrainOrientationFile"];
     InputDataStream.close();
 }
 
-// Ensure that the appropriate files exist - orientation files should be installed, other files should start with the
-// BaseFileName value given from the command line
-void CheckInputFiles(std::string &LogFile, std::string MicrostructureFile, std::string &RotationFilename,
```
streeve, line 79: Same here?

analysis/src/GAutils.cpp
```diff
@@ -46,55 +43,10 @@ void ParseLogFile(std::string LogFile, int &nx, int &ny, int &nz, double &deltax
         std::cout << "Note: orientation filename specified in log file will be used, overriding value from analysis "
                      "input file"
                   << std::endl;
-    RotationFilename = logdata["GrainOrientationFile"];
-    if (RotationFilename.find("GrainOrientationVectors.csv") != std::string::npos) {
```
streeve, line 50: Was this basically moved into the class member functions?

src/CAorientation.hpp
```diff
@@ -0,0 +1,211 @@
...
+        checkFileNotEmpty(rgb_filename);
+
+        // Read and store additional orientation data in the appropriate views
+        view_type_float_host grain_bunge_euler_host = get_orientations_from_file(euler_angles_filename, 3);
+        view_type_float_host grain_rgb_ipfz_host = get_orientations_from_file(rgb_filename, 3);
+
+        // Resize device views and copy data from host
+        Kokkos::realloc(grain_bunge_euler, 3 * n_grain_orientations);
+        Kokkos::realloc(grain_rgb_ipfz, 3 * n_grain_orientations);
+        grain_bunge_euler = Kokkos::create_mirror_view_and_copy(memory_space(), grain_bunge_euler_host);
```
streeve, line 139: Don't think we need to copy if the host was just reallocated

src/CAorientation.hpp
MattRolchigo: The goal here was that the first time an orientations file is read, there's a placeholder value for `n_grain_orientations` that is overwritten. Then, if another orientations file is read, the number of orientations in that file is checked against the already known `n_grain_orientations` to make sure they match. But this can be accomplished in a less convoluted way

analysis/src/GAutils.cpp
MattRolchigo: Yes, this is all in `orientation.get_euler_ipfz_filenames` now

analysis/src/GAutils.cpp
MattRolchigo: Not sure what this comment is referencing, it highlights a deleted line

src/CAorientation.hpp
MattRolchigo: Not sure what the host being reallocated has to do with it. We created host views local to `get_euler_ipfz_filenames` using `get_orientations_from_file` and now we have to copy the data to the device views which are the ones actually stored by the struct

analysis/src/GAutils.cpp
streeve: Sorry, same as in did this get moved into the class

MattRolchigo: Oh, yes it did

src/CAorientation.hpp
streeve: Just seems like a weird place for a comment

src/CAorientation.hpp
streeve: Oh I misread this - why do we bother reallocating when we replace the view on the next line?

MattRolchigo: I was under the (possibly wrong) assumption that the mirror copy only works if the two things are the same size. So I resized the device views so that their size matches the host view size, then copied the host views to the device

streeve: Yeah, you should be able to remove those reallocs without issue

src/CAorientation.hpp
```diff
@@ -0,0 +1,208 @@
...
+    // Using the default exec space for this memory space.
+    using execution_space = typename memory_space::execution_space;
+
+    // Grain orientation data views - either rotation matrices (9 vals per orientation), Euler angles (3 vals per
+    // orientation), or RGB values (3 vals per orientation) Host copy of grain_unit_vector maintained as it will be
+    // needed for all intermediate output steps (host copies of other representations only used in analysis script,
+    // obtained via member functions)
+    int n_grain_orientations;
+    view_type_float grain_unit_vector, grain_bunge_euler, grain_rgb_ipfz;
+    view_type_float_host grain_unit_vector_host;
```
streeve, line 36: Should be able to make this host copy temporary

src/CAorientation.hpp
```diff
@@ -0,0 +1,208 @@
...
+
+    // Using the default exec space for this memory space.
+    using execution_space = typename memory_space::execution_space;
+
+    // Grain orientation data views - either rotation matrices (9 vals per orientation), Euler angles (3 vals per
+    // orientation), or RGB values (3 vals per orientation) Host copy of grain_unit_vector maintained as it will be
+    // needed for all intermediate output steps (host copies of other representations only used in analysis script,
+    // obtained via member functions)
+    int n_grain_orientations;
+    view_type_float grain_unit_vector, grain_bunge_euler, grain_rgb_ipfz;
```
streeve, line 35: According to you, two of these can just be on the host no matter what

src/CAorientation.hpp
```diff
@@ -0,0 +1,196 @@
...
+// By default, start indexing at 0 (GrainID of 1 has Orientation number 0), optionally starting at 1
+// GrainID of 0 is a special case - has orientation 0 no matter what (only used when reconstructing grain ID in
+// getGrainID)
+KOKKOS_INLINE_FUNCTION int get_grain_orientation(const int my_grain_id, const int n_grain_orientations,
+                                                 bool start_at_zero = true) {
+    int my_orientation;
+    if (my_grain_id == 0)
+        my_orientation = 0;
+    else {
+        my_orientation = (abs(my_grain_id) - 1) % n_grain_orientations;
```
streeve, line 180:
```suggestion
        my_orientation = (Kokkos::abs(my_grain_id) - 1) % n_grain_orientations;
```
(also one below)

src/CAorientation.hpp
```diff
@@ -0,0 +1,196 @@
...
+        view_type_float_host grain_unit_vector_host =
+            Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), grain_unit_vector);
+
+        // Find the smallest possible misorientation between the specified direction, and this grain orientations' 6
+        // possible 001 directions (where 62.7 degrees is the largest possible misorientation between two 001 directions
+        // for a cubic crystal system)
+        for (int n = 0; n < n_grain_orientations; n++) {
+            float misorientation_angle_min = 62.7;
+            for (int ll = 0; ll < 3; ll++) {
+                float misorientation = std::abs((180 / M_PI) * acos(grain_unit_vector_host(9 * n + 3 * ll + dir)));
```
streeve, line 150: Inconsistent `std`/missing namepace vs Kokkos math

analysis/src/GArepresentativeregion.hpp
```diff
@@ -824,16 +826,18 @@ struct RepresentativeRegion {
...
                     GrainplotIPF << "0 0 0 0 " << Index1 * deltax * Kokkos::pow(10, 6) << " "
                                  << Index2 * deltax * Kokkos::pow(10, 6) << std::endl;
                 else
-                    GrainplotIPF << GrainEulerAngles(3 * GOVal) << " " << GrainEulerAngles(3 * GOVal + 1) << " "
-                                 << GrainEulerAngles(3 * GOVal + 2) << " 1 " << Index1 * deltax * Kokkos::pow(10, 6)
-                                 << " " << Index2 * deltax * Kokkos::pow(10, 6) << std::endl;
+                    GrainplotIPF << orientation.grain_bunge_euler_host(3 * GOVal) << " "
+                                 << orientation.grain_bunge_euler_host(3 * GOVal + 1) << " "
+                                 << orientation.grain_bunge_euler_host(3 * GOVal + 2) << " 1 "
+                                 << Index1 * deltax * pow(10, 6) << " " << Index2 * deltax * Kokkos::pow(10, 6)
```
streeve, line 839: :)

analysis/src/GArepresentativeregion.hpp
```diff
@@ -824,16 +826,18 @@ struct RepresentativeRegion {
                     throw std::runtime_error("Error: Unknown regionOrientation input for WriteIPFColoredCrossSection: "
                                              "should be XY, YZ, or XZ");
                 // What orientation does this grain id correspond to? Should be between 0 and NumberOfOrientations-1
-                int GOVal = (abs(GrainID(ZLoc, XLoc, YLoc)) - 1) % NumberOfOrientations;
+                int GOVal = (abs(GrainID(ZLoc, XLoc, YLoc)) - 1) % orientation.n_grain_orientations;
```
streeve, line 829: Looks like my mistake on the missing namespace here

streeve (approved):

PR #272, opened 2023-12-19
Use Kokkos math (require Kokkos 4.0)

streeve:
Needed for kernel portability, but used on the host as well


streeve:
Apparently there is no `hypot` float in 3.7

MattRolchigo (approved):

streeve:
Updated the release notes

PR #275, opened 2024-01-05
Enable commandline args for unit tests

### LINKED ISSUE #67: Unit test for InputReadFromFile must be run from same directory as executable
No paths are currently given for the two examples files - Inp_DirSolidification.txt and Inp_SpotMelt.txt - that were installed in the build directory for unit testing. The unit test for InputReadFromFile must currently be run from the same directory as the unit test executable, this should be remedied by providing a path to these files

streeve:
Needed outside of using ctest directly. Not a fan of the global variable, but it works. Closes #67

### DIFF
--- a/unit_test/tstInputs.hpp
+++ b/unit_test/tstInputs.hpp
@@ -13,14 +13,24 @@
 #include <string>
 #include <vector>
 
+#include <unit_test_cmd_line.hpp>
+
 namespace Test {
-//---------------------------------------------------------------------------//
-// file_read_tests
-//---------------------------------------------------------------------------//
-void WriteTestData(std::string InputFilename, int PrintVersion) {
+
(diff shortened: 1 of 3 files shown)


MattRolchigo:
What do you mean by "outside of running ctest directly"? And what exactly is the command line argument here, the path to the /build/unit_test directory?

streeve:
> What do you mean by "outside of running ctest directly"? And what exactly is the command line argument here, the path to the /build/unit_test directory?

If you run using `ctest` it can find the copied input files just fine. But now you can pass the path to either `examples/` or copied files in `build/unit_test` (or anywhere really) to `ExaCA_Inputs_test_*`

MattRolchigo (approved):

streeve:
Closing - I don't think this is worth the extra code

PR #278, opened 2024-01-05
Convert c-style casts to static_cast

### LINKED ISSUE #152: Data type consistency
Use of `int`, `float`, `long int`, and `double` is not consistent from function to function, with hidden casts and c-style casts between types

streeve:
Part of #152


MattRolchigo:
I don't know if this alone should be enough to close #152, I'd like to go file by file and make sure everything is stored as a consistent type without any hidden casts first

streeve:
> I don't know if this alone should be enough to close #152, I'd like to go file by file and make sure everything is stored as a consistent type without any hidden casts first

Modified so this won't close the issue yet. Sounds like your changes can be done separately though

MattRolchigo (approved):

PR #282, opened 2024-01-12
Combine initialize routines and fix init undercooling bug

MattRolchigo:
- Added additional constraints on `G`, `R`, and `InitUndercooling` inputs for problem types `C` and `SingleGrain`
- Consolidated temperature initialize routines for cases with uniform cooling and `G` given as either zero or non-zero
- Fixed bug in initial undercooling field for problem type `C` when a non-zero value is given for `InitUndercooling`
- Updated temperature unit test and added `G` as a test input

### DIFF
--- a/src/CAinputs.hpp
+++ b/src/CAinputs.hpp
@@ -234,13 +234,13 @@ struct Inputs {
             // problem type C if not given, should be a positive number
             if (inputdata["TemperatureData"].contains("InitUndercooling"))
                 if (inputdata["TemperatureData"]["InitUndercooling"] < 0)
-                    throw std::runtime_error("Error: optional temperature data argument `InitUndercooling` should be "
+                    throw std::runtime_error("Error: optional temperature data argument InitUndercooling should be "
                                              "greater than or equal to zero");
             if (SimulationType == "SingleGrain")
                 temperature.initUndercooling = inputdata["TemperatureData"]["InitUndercooling"];
             else if ((SimulationType == "C") && (inputdata["TemperatureData"].contains("InitUndercooling")))
                 temperature.initUndercooling = inputdata["TemperatureData"]["InitUndercooling"];
-            if ((temperature.G > 0) && (temperature.R == 0)) {
+            if ((temperature.G > 0) && (fabs(temperature.R) < 0.000001)) {
                 // Throw error for edge case where the cooling rate is 0, but cells in the domain would be initialized
                 // above the liquidus temperature (i.e., cells that would never solidify)
                 int location_init_undercooling;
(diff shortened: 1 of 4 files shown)


unit_test/tstTemperature.hpp
```diff
@@ -308,9 +314,10 @@ TEST(TEST_CATEGORY, temperature) {
...
     }
     // Test for directional and single grain problems, and with and without a thermal gradient for the single grain
-    // problem
-    testInit_UnidirectionalGradient("C", 1000000);
-    testInit_UnidirectionalGradient("SingleGrain", 0);
-    testInit_UnidirectionalGradient("SingleGrain", 1000000);
+    // problem and with/without an initial undercooling present at the initial solidification front
+    testInit_UnidirectionalGradient("C", 1000000, 0);
+    testInit_UnidirectionalGradient("C", 1000000, 2);
+    //    testInit_UnidirectionalGradient("SingleGrain", 0, 2);
```
streeve, line 320: Do you want to keep these?

src/CAinputs.hpp
```diff
@@ -231,11 +231,30 @@ struct Inputs {
             temperature.G = inputdata["TemperatureData"]["G"];
             temperature.R = inputdata["TemperatureData"]["R"];
             // Optional initial undercooling for problem type C (required for type SingleGrain). Defaults to 0 for
-            // problem type C if not given
+            // problem type C if not given, should be a positive number
+            if (inputdata["TemperatureData"].contains("InitUndercooling"))
+                if (inputdata["TemperatureData"]["InitUndercooling"] < 0)
+                    throw std::runtime_error("Error: optional temperature data argument `InitUndercooling` should be "
```
streeve, line 237: I don't think the single quotes are needed

streeve (commented):
Looks good overall, couple small things

src/CAinputs.hpp
```diff
@@ -234,13 +234,13 @@ struct Inputs {
...
                 if (inputdata["TemperatureData"]["InitUndercooling"] < 0)
-                    throw std::runtime_error("Error: optional temperature data argument `InitUndercooling` should be "
+                    throw std::runtime_error("Error: optional temperature data argument InitUndercooling should be "
                                              "greater than or equal to zero");
             if (SimulationType == "SingleGrain")
                 temperature.initUndercooling = inputdata["TemperatureData"]["InitUndercooling"];
             else if ((SimulationType == "C") && (inputdata["TemperatureData"].contains("InitUndercooling")))
                 temperature.initUndercooling = inputdata["TemperatureData"]["InitUndercooling"];
-            if ((temperature.G > 0) && (temperature.R == 0)) {
+            if ((temperature.G > 0) && (fabs(temperature.R) < 0.000001)) {
```
streeve, line 243: Kokkos or std namespace

streeve (approved):
I know you're just trying to do it incrementally, but all the variable name changes in the same commit as real changes make it hard to review

MattRolchigo:
I promise after these 2 merges I'll go through and fix the function/variable names everywhere so we don't have to worry about that anymore

PR #283, opened 2024-01-17
Reorganized functions in CAparsefiles

MattRolchigo:
Parsing functions exclusively used as part of the inputs, grid, or temperature struct have been moved into the appropriate file. Parsing functions used in multiple points in the code (or will be in the future, i.e. reading vtk field data as part of restarting ExaCA from a previous state) have been consolidated in `CAparsefiles.cpp` and `CAparsefiles.hpp`

### DIFF
--- a/src/CAgrid.hpp
+++ b/src/CAgrid.hpp
@@ -151,6 +151,89 @@ struct Grid {
                       << " total cells in the Z direction" << std::endl;
     };
 
+    // Read x, y, z coordinates in tempfile_thislayer (temperature file in either an ASCII or binary format) and return
+    // the min and max values. This is not used by the temperature struct (no temperature info is parsed or stored), and
+    // is only used to find the bounds of the simulation domain
+    std::array<double, 6> parseTemperatureCoordinateMinMax(std::string tempfile_thislayer, bool BinaryInputData) {
+
+        std::array<double, 6> XYZMinMax;
+        std::ifstream TemperatureFilestream;
+        TemperatureFilestream.open(tempfile_thislayer);
+        std::size_t vals_per_line;
+
+        // Binary temperature data should contain only the six columns of interest
+        // Comma-separated double type values may contain additional columns after the 6 used by ExaCA
+        if (BinaryInputData)
+            vals_per_line = 6;
+        else {
+            // Read the header line data
+            // Make sure the first line contains all required column names: x, y, z, tm, tl, cr
+            std::string HeaderLine;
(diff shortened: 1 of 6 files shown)


src/CAgrid.hpp
```diff
@@ -151,6 +151,89 @@ struct Grid {
                       << " total cells in the Z direction" << std::endl;
     };
 
+    // Read x, y, z coordinates in tempfile_thislayer (temperature file in either an ASCII or binary format) and return
+    // the min and max values. This is not used by the temperature struct (no temperature info is parsed or stored), and
+    // is only used to find the bounds of the simulation domain
+    std::array<double, 6> parseTemperatureCoordinateMinMax(std::string tempfile_thislayer, bool BinaryInputData) {
```
streeve, line 157: Why not in `Temperature`?

src/CAparsefiles.cpp
```diff
@@ -207,84 +142,33 @@ bool checkTemperatureFileFormat(std::string tempfile_thislayer) {
     return BinaryInputData;
 }
 
-// Read x, y, z coordinates in tempfile_thislayer (temperature file in either an ASCII or binary format) and return the
-// min and max values
-std::array<double, 6> parseTemperatureCoordinateMinMax(std::string tempfile_thislayer, bool BinaryInputData) {
+// Check to make sure that the 6 expected column names appear in the correct order in the header for this temperature
+// file Return the number of columns present - ignore any columns after the 6 of interest
+std::size_t checkForHeaderValues(std::string header_line) {
```
streeve, line 147: Seems very temperature specific..

src/CAgrid.hpp
MattRolchigo: I knew you'd ask that which is why I added the comment on line 155: "This is not used by the temperature struct (no temperature info is parsed or stored), and is only used to find the bounds of the simulation domain". I thought about renaming it to something without "Temperature" in the name (parseCoordinateMinMaxFromFiles?), but I wanted to keep this PR to as few changes outside of pure reorganization as possible

src/CAparsefiles.cpp
MattRolchigo: Same thing as the other one, its called from both `grid` in the `parseTemperatureCoordMinMax` function and from `temperature` itself

src/CAgrid.hpp
streeve: You got me there..

streeve (approved):

PR #294, opened 2024-02-08
Print initialized ny_local and y_offset in rank order

MattRolchigo:
Previously, all ranks would attempt to concurrently print this information to std::cout, sometimes overwriting each other. Rank 0 now prints all size and offset data to avoid this

### DIFF
--- a/src/CAgrid.hpp
+++ b/src/CAgrid.hpp
@@ -134,7 +134,16 @@ struct Grid {
 
         // Add halo regions with a width of 1 in +/- Y if this MPI rank is not as a domain boundary in said direction
         addHalo();
-        std::cout << "Rank " << id << " spans Y = " << y_offset << " through " << y_offset + ny_local - 1 << std::endl;
+        // Gather ny_local and y_offset information on rank 0 to print to screen in rank order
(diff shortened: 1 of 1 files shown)


src/CAgrid.hpp
```diff
@@ -134,7 +134,16 @@ struct Grid {
 
         // Add halo regions with a width of 1 in +/- Y if this MPI rank is not as a domain boundary in said direction
         addHalo();
-        std::cout << "Rank " << id << " spans Y = " << y_offset << " through " << y_offset + ny_local - 1 << std::endl;
+        // Gather ny_local and y_offset information on rank 0 to print to screen in rank order
+        int *recvbuf_offset = (int *)malloc(np * sizeof(int));
```
streeve, line 138: Can we use truly any other way to allocate an array?

MattRolchigo: Probably? When it comes to C array stuff with MPI I sometimes just default to whatever I got used to doing when I first learned MPI. As long as it doesn't break anything I'm happy to change that line

MattRolchigo:
:( My beautiful C arrays

streeve (approved):

PR #303, opened 2024-02-22
Fix error in misorientation calculation

MattRolchigo:
Correct the calculation for the misalignment between a grain's <100> crystallographic directions and a given cardinal direction (previously only compared the misalignment for 3 of the 6 <100> directions, and incorrectly stated a max misorientation value in the comment)


streeve:
Also this sounds like something to add to the release notes

PR #307, opened 2024-03-06
Rewrite timing class as  seperate timers

streeve:


MattRolchigo (approved):
Thank you for doing this, it annoyed me forever how much repeated code I had in these lines

streeve:
> Thank you for doing this, it annoyed me forever how much repeated code I had in these lines

That's what a conference is good for :)

PR #308, opened 2024-03-08
Rename libraries and executables

### LINKED ISSUE #309: Changing executable and library names
Planned changes:

- `ExaCA-Kokkos`->`ExaCA` 
  - Intended to clarify that this is the primary executable
  - symlink to old exe name
- `ExaCA::ExaCA`->`ExaCA::Core`
  - no planned alias in order to rename ExaCA exe (cannot share name)
- `grain_analysis`->`ExaCA-GrainAnalysis`
  - Intended to clarify this is part of ExaCA
  - symlink to old exe name
- `ExaCA::ExaGA`->`ExaCA::Analysis`
  - no planned alias

Please comment with any anticipated issues with this renaming and linking

streeve:
Renames exe, but keeps symlinks for compatibility for now. Can choose different names if we want of course

Renames libraries without alias to old names

Closes #309

### DIFF
--- a/analysis/bin/CMakeLists.txt
+++ b/analysis/bin/CMakeLists.txt
@@ -1,3 +1,9 @@
-add_executable(grain_analysis runGA.cpp)
-target_link_libraries(grain_analysis ExaGA)
-install(TARGETS grain_analysis DESTINATION ${CMAKE_INSTALL_BINDIR})
+add_executable(ExaCA-GrainAnalysis runGA.cpp)
+target_link_libraries(ExaCA-GrainAnalysis ExaCA-Analysis)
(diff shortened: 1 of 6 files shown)


MattRolchigo (approved):

streeve:
> I don't know how I feel about ExaCA-Core as the new name of the executable, but I'll get used to it eventually

That's my proposed new library name, with the new exe `ExaCA`. If we want to rename all the libs and aliases I need to update with new aliases for backwards compatibility

MattRolchigo:
I confirmed that I could compile your library/executable rename branch on AMD, and that I could run with `ExaCA-Kokkos` or `ExaCA` as the executable name (same thing with analysis with `grain_analysis` or `ExaCA-GrainAnalysis`). The only thing I'd suggest is updating the README letting the users know that these were renamed and that presumably at some point in the future, the symlinks will be removed

PR #316, opened 2024-03-25
Option to record solidification front undercooling for problem type C

MattRolchigo:
Optionally prints a separate file consisting of "Z (µm), undercooling start, undercooling end" values for the leading edge of the solidification front for directional solidification simulations

### DIFF
--- a/src/CAprint.hpp
+++ b/src/CAprint.hpp
@@ -337,6 +337,9 @@ struct Print {
                 }
                 interlayer_all_layers_ofstream.close();
             }
+            // File of front undercooling at interfa
+            if (_inputs.print_front_undercooling)
+                temperature.writeFrontUndercooling(path_base_filename, id, grid);
 
             // Views where data should be printed only for the layer of the problem that just finished
             if (_inputs.interlayer_current) {
@@ -515,6 +518,20 @@ struct Print {
             if (!(_inputs.print_binary))
                 output_fstream << std::endl;
         }
+    }
+
+    // Called on rank 0 to write solidification undercooling data to a file
+    template <typename Print1DViewType>
+    void printSolidificationFrontUndercooling(const int nz, Print1DViewType front_undercooling) {
+
+        std::string front_undercooling_filename = path_base_filename + "_FrontUndercooling.csv";
+        std::ofstream und_ofstream;
+        und_ofstream.open(front_undercooling_filename);
+        for (int coord_z = 0; coord_z < nz - 1; coord_z++) {
+            und_ofstream << front_undercooling(coord_z) << std::endl;
+        }
+        und_ofstream << front_undercooling(nz - 1);
+        und_ofstream.close();
     }
 
     // On rank 0, print grain misorientation, 0-62 for epitaxial grains and 100-162 for nucleated grains, to a vtk
--- a/src/CAprint.hpp
+++ b/src/CAprint.hpp
@@ -530,8 +530,8 @@ struct Print {
         und_ofstream.open(front_undercooling_filename);
         und_ofstream << "Z (micrometers), Initial Undercooling, Final Undercooling" << std::endl;
         for (int coord_z = 0; coord_z < nz - 1; coord_z++) {
-            und_ofstream << coord_z * deltax * pow(10, 6) << "," << front_undercooling(coord_z, 0) << ","
-                         << front_undercooling(coord_z, 1) << std::endl;
+            und_ofstream << static_cast<double>(coord_z) * deltax * pow(10, 6) << "," << front_undercooling(coord_z, 0)
+                         << "," << front_undercooling(coord_z, 1) << std::endl;
         }
         und_ofstream << static_cast<double>(nz - 1) * deltax * pow(10, 6) << "," << front_undercooling(nz - 1, 0) << ","
                      << front_undercooling(nz - 1, 1);
--- a/src/CAprint.hpp
+++ b/src/CAprint.hpp
@@ -337,10 +337,11 @@ struct Print {
                 }
                 interlayer_all_layers_ofstream.close();
             }
-            // File of front undercooling at interfa
-            if (_inputs.print_front_undercooling)
-                temperature.writeFrontUndercooling(path_base_filename, id, grid);
-
+            // File of front undercooling at interface
+            if (_inputs.print_front_undercooling) {
+                auto start_end_solidification_z = temperature.getFrontUndercoolingStartFinish(id, grid);
+                printSolidificationFrontUndercooling(grid.deltax, grid.nz, start_end_solidification_z);
+            }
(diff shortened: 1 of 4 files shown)


src/CAtemperature.hpp
```diff
@@ -664,6 +665,56 @@ struct Temperature {
         undercooling_solidification_start = Kokkos::subview(undercooling_solidification_start_all_layers, layer_range);
     }
 
+    // For each Z coordinate, find the smallest undercooling at which solidification started and finished, writing this
+    // data to an output file
+    void writeFrontUndercooling(std::string path_base_filename, const int id, const Grid &grid) {
```
streeve, line 670: I'm sure you thought about this, but does part of this belong more in `Print` (at least the file output portion)?

src/CAprint.hpp
```diff
@@ -517,6 +520,20 @@ struct Print {
...
+    template <typename Print1DViewType>
+    void printSolidificationFrontUndercooling(const int nz, Print1DViewType front_undercooling) {
+
+        std::string front_undercooling_filename = path_base_filename + "_FrontUndercooling.csv";
+        std::ofstream und_ofstream;
+        und_ofstream.open(front_undercooling_filename);
+        for (int coord_z = 0; coord_z < nz - 1; coord_z++) {
+            und_ofstream << front_undercooling(coord_z) << std::endl;
+        }
+        und_ofstream << front_undercooling(nz - 1);
```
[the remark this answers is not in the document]
MattRolchigo: Probably not, just figured that not having a blank line at the end of the file could save some headaches later

src/CAtemperature.hpp
MattRolchigo: I think I might have been in the process of making that change given the unused `printSolidificationFrontUndercooling` function in `print`

src/CAprint.hpp
```diff
@@ -517,6 +521,23 @@ struct Print {
...
+    // Called on rank 0 to write solidification undercooling data to a file
+    template <typename Print2DViewType>
+    void printSolidificationFrontUndercooling(const double deltax, const int nz, Print2DViewType front_undercooling) {
+
+        std::string front_undercooling_filename = path_base_filename + "_FrontUndercooling.csv";
+        std::ofstream und_ofstream;
+        und_ofstream.open(front_undercooling_filename);
+        und_ofstream << "Z (micrometers), Initial Undercooling, Final Undercooling" << std::endl;
+        for (int coord_z = 0; coord_z < nz - 1; coord_z++) {
+            und_ofstream << coord_z * deltax * pow(10, 6) << "," << front_undercooling(coord_z, 0) << ","
```
streeve, line 533: No `static_cast` here?

src/CAtemperature.hpp
```diff
@@ -604,6 +607,57 @@ struct Temperature {
...
+    // For each Z coordinate, find the smallest undercooling at which solidification started and finished, writing this
+    // data to an output file
+    view_type_float_2d_host getFrontUndercoolingStartFinish(const int id, const Grid &grid) {
+        view_type_float start_solidification_z(Kokkos::ViewAllocateWithoutInitializing("start_solidification_z"),
+                                               grid.nz);
+        view_type_float end_solidification_z(Kokkos::ViewAllocateWithoutInitializing("end_solidification_z"), grid.nz);
+        auto _undercooling_solidification_start = undercooling_solidification_start;
+        auto _undercooling_current = undercooling_current;
+        Kokkos::parallel_for(
+            "GetMinUndercooling", grid.nz, KOKKOS_LAMBDA(const int &coord_z) {
```
streeve, line 619: I think you need a policy here

MattRolchigo: Do I though? I'm just looping 0 through nz-1, kind of like the parallel_for loops in CAupdate.hpp which don't have policies

src/CAtemperature.hpp
```diff
@@ -604,6 +607,57 @@ struct Temperature {
...
+                start_solidification_z(coord_z) = min_start_undercooling;
+                end_solidification_z(coord_z) = min_end_undercooling;
+            });
+
+        // Rank 0 - collect min values from all ranks to get the global start/end solidification undercoolings
+        view_type_float start_solidification_z_reduced(
+            Kokkos::ViewAllocateWithoutInitializing("start_solidification_z_red"), grid.nz);
+        view_type_float end_solidification_z_reduced(
+            Kokkos::ViewAllocateWithoutInitializing("end_solidification_z_red"), grid.nz);
+        MPI_Reduce(start_solidification_z.data(), start_solidification_z_reduced.data(), grid.nz, MPI_FLOAT, MPI_MIN, 0,
```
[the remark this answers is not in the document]
MattRolchigo: Given the MPI issues I had with 2D views and accessing the data, I figured this was safer. There may be a safe way to do it on the 2D views

src/CAtemperature.hpp
streeve: Anytime we don't use a policy it uses the default and hides the fact that it only works because we use the default space in `runCA` - I forgot we were still doing that in some places

src/CAtemperature.hpp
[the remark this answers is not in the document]
streeve: Fair point - I'll take a look, but if we keep it as is we should add a comment about that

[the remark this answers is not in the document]
MattRolchigo: There's also the fact that if I did the reductions on the 2D views, I'd need to do one reduction on one column (the one holding the `start_solidification_z` data) and one on the other (the one holding the `end_solidification_z` data). Depending on how the data is stored in memory, a mistake could easily be possible there.

src/CAtemperature.hpp
```diff
@@ -604,6 +607,60 @@ struct Temperature {
...
+                        if (_undercooling_current(index) < min_end_undercooling)
+                            min_end_undercooling = _undercooling_current(index);
+                    }
+                }
+                start_solidification_z(coord_z) = min_start_undercooling;
+                end_solidification_z(coord_z) = min_end_undercooling;
+            });
+
+        // Rank 0 - collect min values from all ranks to get the global start/end solidification undercoolings
+        view_type_float start_solidification_z_reduced(
```
streeve, line 637: Can we do this in place?

src/CAprint.hpp
```diff
@@ -517,6 +521,23 @@ struct Print {
         }
     }
 
+    // Called on rank 0 to write solidification undercooling data to a file
```
streeve, line 524: I don't see how this is only called on rank 0

MattRolchigo: You're right - fixed

src/CAtemperature.hpp
MattRolchigo: Apparently not

streeve (approved):

PR #318, opened 2024-03-26
Additional customization of problem type C substrate init

MattRolchigo:
Options to initialize bottom surface for problem type C using either
- Fraction of sites active: guarantees that each grain is assigned a cell, but the grain locations will change if the domain size is changed) 
Or
- Density of grains: guarantees that grains are placed in the same locations even if the cell size is changed as long as the bottom surface size and proportions remain the same, but some grains will not be assigned a cell if the density is too large to resolve for the given cell size

### DIFF
--- a/examples/Inp_SmallDirSolidification.json
+++ b/examples/Inp_SmallDirSolidification.json
@@ -20,7 +20,7 @@
       "R": 300000
    },
    "Substrate": {
-      "FractionSurfaceSitesActive": 0.25
+      "SurfaceSiteDensity": 0.25
    },
    "Printing": {
       "PathToOutput": "./",
--- a/examples/README.md
+++ b/examples/README.md
@@ -90,10 +90,11 @@ The .json files in the examples subdirectory are provided on the command line to
 ## Substrate inputs
 | Input        | Relevant problem type(s))| Details |
 |--------------| -------------------------|---------|
-|FractionSurfaceSitesActive | C           | What fraction of cells at the bottom surface of the domain are the source of a grain? (see note (b))
-|GrainLocationsX | C           | List of grain locations in X on the bottom surface of the domain (see note (b))
-|GrainLocationsY | C           | List of grain locations in Y on the bottom surface of the domain (see note (b))
-|GrainIDs | C           | GrainID values for each grain in (X,Y) (see note (b))
+|FractionSurfaceSitesActive | C           | What fraction of cells at the bottom surface of the domain are the source of a grain? (see note (b-i))
+|SurfaceSiteDensity         | C           | Density, in nuclei/µm^2, of grains at the bottom surface of the domain (see note (b-ii))
+|GrainLocationsX | C           | List of grain locations in X on the bottom surface of the domain (see note (b-iii))
+|GrainLocationsY | C           | List of grain locations in Y on the bottom surface of the domain (see note (b-iii))
+|GrainIDs | C           | GrainID values for each grain in (X,Y) (see note (b3))
 |FillBottomSurface | C  | Optionally assign all cells on the bottom surface the grain ID of the closest grain (defaults to false)
 |MeanSize      | Spot, R                  | Mean spacing between grain centers in the baseplate/substrate (in microns) (see note (a))
 |SubstrateFilename |  Spot, R             | Path to and filename for substrate data (see note (a))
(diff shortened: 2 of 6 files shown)


examples/Inp_SmallDirSolidification.json
```diff
@@ -20,7 +20,7 @@
       "R": 300000
    },
    "Substrate": {
-      "FractionSurfaceSitesActive": 0.25
+      "SurfaceSiteDensity": 0.25
```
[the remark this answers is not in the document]
MattRolchigo: This was changed to make sure the problem result is the same. Because the "FractionSurfaceSitesActive" mode of initializing the problem is now the "SurfaceSiteDensity" mode (with new code written for the "FractionSurfaceSitesActive" mode), I had to change the input type. Because of the cell size of 1, the number didn't change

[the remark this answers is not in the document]
streeve: Makes sense, but the names don't make that very clear. I'll try to think of something..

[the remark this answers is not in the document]
MattRolchigo: The real problem is that the `FractionSurfaceSitesActive` wasn't really setting the fraction before, it was selecting a number of x and y coordinates corresponding to FractionSurfaceSitesActive * NumberPossibleCells. Since only one grain can exist at a given (x,y), `FractionSurfaceSitesActive` ended up underestimating the actual fraction.

unit_test/tstCellData.hpp
```diff
@@ -92,11 +92,20 @@ void testCellDataInit_ConstrainedGrowthMultiGrain() {
...
     Inputs inputs;
     Grid grid;
+    inputs.substrate.surface_init_mode = input_surface_init_mode;
+    if (input_surface_init_mode == "SurfaceSiteDensity") {
+        // Set so that there are 4 grains in the domain, regardless of domain size
+        inputs.substrate.surface_site_density = 1.0 / (4.0 * static_cast<double>(np));
+    }
+    else {
+        // Set half of sites active, number of grains will depend on domain size
+        inputs.substrate.fract_surface_sites_active = 0.25;
```
streeve, line 102: Should be 0.5?

MattRolchigo: No the comment was wrong, the test previously set half of them active but here I wanted the tests to line up so that we'd get the same number of grains regardless of which way the init was performed

examples/README.md
```diff
@@ -90,10 +90,11 @@ The .json files in the examples subdirectory are provided on the command line to
 ## Substrate inputs
 | Input        | Relevant problem type(s))| Details |
 |--------------| -------------------------|---------|
-|FractionSurfaceSitesActive | C           | What fraction of cells at the bottom surface of the domain are the source of a grain? (see note (b))
-|GrainLocationsX | C           | List of grain locations in X on the bottom surface of the domain (see note (b))
-|GrainLocationsY | C           | List of grain locations in Y on the bottom surface of the domain (see note (b))
-|GrainIDs | C           | GrainID values for each grain in (X,Y) (see note (b))
+|FractionSurfaceSitesActive | C           | What fraction of cells at the bottom surface of the domain are the source of a grain? (see note (b-i))
+|SurfaceSiteDensity         | C           | Density, in nuclei/µm^2, of grains at the bottom surface of the domain (see note (b-ii))
```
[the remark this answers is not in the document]
streeve: Should add a note for the deprecated `FractionSurfaceSitesActive`

unit_test/tstCellData.hpp
```diff
@@ -435,8 +448,9 @@ void testCalcVolFractionNucleated() {
 //---------------------------------------------------------------------------//
 TEST(TEST_CATEGORY, cell_init_tests) {
     testCellDataInit_SingleGrain();
-    testCellDataInit_ConstrainedGrowthMultiGrain();
-    testCellDataInit_ConstrainedGrowthTwoGrain();
+    testCellDataInit_Constrained_Automatic("SurfaceSiteDensity");
+    testCellDataInit_Constrained_Automatic("FractionSurfaceSitesActive");
```
streeve, line 452: Deprecated input

MattRolchigo: That one I actually left on purpose to make sure it still works with the deprecated input, but point taken it should also test "SurfaceSiteFraction"

MattRolchigo: In the past, I don't think I've noted deprecated inputs in the README files. I think that just makes them more cluttered, if a user really wants to know what an input is then they should use the up-to-date ones

streeve (approved):

PR #319, opened 2024-04-02
Option for custom initial octahedron size

MattRolchigo:
Defaults to the previously used value of 0.01 - but the option to select an initial size allows for better comparisons when changing cell sizes (for example, a size of 0.01 could be used with a cell size of 1 µm, and a size of 0.005 could be used with a cell size of 2 µm)

### DIFF
--- a/src/CAinterface.hpp
+++ b/src/CAinterface.hpp
@@ -43,6 +43,8 @@ struct Interface {
     view_type_buffer buffer_south_send, buffer_north_send, buffer_south_recv, buffer_north_recv;
     view_type_int send_size_south, send_size_north, steering_vector, num_steer;
     view_type_int_host send_size_south_host, send_size_north_host, num_steer_host;
+    // Initial size of new octahedra
+    float _init_oct_size;
 
     // Neighbor lists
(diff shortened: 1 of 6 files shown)


src/CAinterface.hpp
```diff
@@ -188,7 +191,7 @@ struct Interface {
     KOKKOS_INLINE_FUNCTION
     void createNewOctahedron(const int index, const int coord_x, const int coord_y, const int y_offset,
                              const int coord_z) const {
-        diagonal_length(index) = 0.01;
+        diagonal_length(index) = _init_oct_size;
```
streeve, line 194: This is fine because you capture this object locally and then access it without nested classes, right?

streeve (approved):

src/CAinterface.hpp
MattRolchigo: To be honest, I haven't tested this on GPU yet so I'm not sure. Based on the fact that it's been fine with other members of the `interface` class accessed in these inline functions (which are called from CAupdate.hpp subroutines, not from member functions of `interface` itself), I expect/hope that you're right

MattRolchigo: Okay it works on GPU

PR #321, opened 2024-04-05
Reorganize intermediate output to console

MattRolchigo:
Print all information that is collected during intermediate output subroutine, and write to the console in a more concise way using multiple lines (and a separator between output steps)


streeve (approved):
But what if I want a wall of text instead?

MattRolchigo:
A wall of text would've probably been an improvement over the super long single line that doesn't fit in the width of a terminal

PR #322, opened 2024-04-09
Replace custom json output with `json.dump()`

streeve:


src/CAinputs.hpp
```diff
@@ -644,90 +646,82 @@ struct Inputs {
                        const double x_min, const double x_max, const double y_min, const double y_max,
                        const double z_min, const double z_max, const float vol_fraction_nucleated) {
 
-        int *ny_local_allranks = new int[np];
-        int *y_offset_allranks = new int[np];
```
MattRolchigo, line 648: :( My C arrays...

src/CAinputs.hpp
```diff
@@ -644,90 +646,82 @@ struct Inputs {
...
+            json_log["TimeStepOfOutput"] = cycle;
+            json_log["SimulationType"] = simulation_type;
+            json_log["GrainOrientationFile"] = grain_orientation_file;
+            json_log["Domain"] = {};
+            json_log["Domain"]["Nx"] = nx;
+            json_log["Domain"]["Ny"] = ny;
+            json_log["Domain"]["Nz"] = nz;
+            json_log["Domain"]["CellSize"] = deltax;
+            json_log["Domain"]["Timestep"] = domain.deltat;
+            std::array<double, 2> x_bounds = {x_min, x_max};
```
MattRolchigo, line 670: It looks like this will print the two bounds on separate lines, rather than [low,high] which looks kind of clunky... is there a way to change that?

src/CAinputs.hpp
```diff
@@ -644,90 +646,82 @@ struct Inputs {
...
-            exaca_log << "      \"Ny\": " << ny << "," << std::endl;
-            exaca_log << "      \"Nz\": " << nz << "," << std::endl;
-            exaca_log << "      \"CellSize\": " << deltax << "," << std::endl;
-            exaca_log << "      \"TimeStep\": " << domain.deltat << "," << std::endl;
-            exaca_log << "      \"XBounds\": [" << x_min << "," << x_max << "]," << std::endl;
-            exaca_log << "      \"YBounds\": [" << y_min << "," << y_max << "]," << std::endl;
-            exaca_log << "      \"ZBounds\": [" << z_min << "," << z_max << "]";
+            // Build json output.
+            nlohmann::json json_log;
+            json_log["ExaCAVersion"] = version();
```
MattRolchigo, line 657: I know I'm being picky here, but it looks like the json_log orders the outputs in the dump in alphabetical order, so unlike before where the commit hash and version would appear at the top of the log file, they're now in the middle. Is that customizable?

src/CAinterfacialresponse.hpp
```diff
@@ -118,19 +118,16 @@ struct InterfacialResponseFunction {
...
-            << "\"" << functionName() << "\"," << std::endl;
-        out << "       \"A\": " << (A) << "," << std::endl;
-        out << "       \"B\": " << (B) << "," << std::endl;
-        out << "       \"C\": " << (C) << "," << std::endl;
+    auto print() {
+        nlohmann::json log;
+        log["Function"] = functionName();
+        log["A"] = A;
+        log["B"] = B;
+        log["C"] = C;
```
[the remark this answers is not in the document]
streeve: Maybe possible, but not sure how

src/CAinputs.hpp
streeve: I think the only thing we could do is add more sections and accept those in alphabetical-key order

MattRolchigo:
I think we should keep the json output the way it is, the format is cleaner and easier to read and it doesn't sound like the dump has a lot of customizability

PR #323, opened 2024-04-09
Rename problems for directional solidification and temperature histories from files

streeve:


analysis/src/GAutils.cpp
```diff
@@ -35,8 +35,13 @@ void parseLogFile(std::string log_file, int &nx, int &ny, int &nz, double &delta
     ny = logdata["Domain"]["Ny"];
     nz = logdata["Domain"]["Nz"];
     std::string simulation_type = logdata["SimulationType"];
-    if (simulation_type == "C")
+    if (simulation_type == "DirSol" || simulation_type == "C") {
```
MattRolchigo, line 38: How would you feel about "Directional" as the problem type? While we both know what DirSol is, I feel like the names of the other problem types describe the problems better ("FromFile", "FromFinch", "Spot", "SingleGrain")

src/CAinputs.hpp
```diff
@@ -160,23 +160,26 @@ struct Inputs {
...
-        // DirSoldification problems now include remelting logic
-        if (simulation_type == "RM")
-            simulation_type = "R";
-        else if (simulation_type == "R")
-            std::cout << "Warning: While the specified problem type did not include remelting, all simulations now "
-                         "include remelting"
-                      << std::endl;
-        else if ((simulation_type == "S") || (simulation_type == "SM"))
+        // "FromFile": time-temperature history comes from external files
+        if (simulation_type == "CM" || simulation_type == "C") {
```
MattRolchigo, line 166: There's never been a "CM" problem type, "C" never had remelting

MattRolchigo (approved):

MattRolchigo:
@streeve I am now

PR #330, opened 2024-04-23
Update copyright years

MattRolchigo:


streeve:
Not doing this is en vogue (e.g. https://nexb.com/do-you-really-need-to-update-the-copyright-each-new-year/). We could 

1. remove the years
2. revert back to just the original year
3. merge this changeset as is

streeve (changes requested):
Let's update the license file to just say 2021, then I'm good

streeve (approved):

PR #331, opened 2024-04-23
fixup: Reset counter to print time for each layer

MattRolchigo:
Only the time from the previous layer (not the cumulative runtime of all layers) is printed at the conclusion of said layer for multilayer problems. Fixup of #298


streeve (approved):
I made a lot of mistakes in that PR..

PR #333, opened 2024-04-24
Outer steering vector

MattRolchigo:
Slight performance optimization for problem types with remelting to iterate over an "outer" steering vector (created at a time interval of either 5000 time steps or a user input value) rather than over all cells when creating the main steering vector used for cell capture.

### DIFF
--- a/src/CAinputs.hpp
+++ b/src/CAinputs.hpp
@@ -26,6 +26,7 @@ struct Inputs {
 
     std::string simulation_type = "", material_filename = "", grain_orientation_file = "";
     unsigned long rng_seed = 0.0;
+    int build_increment_outer = 5000;
     DomainInputs domain;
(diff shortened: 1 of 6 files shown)


streeve (changes requested):
Should add to the readme. Maybe add this to one of the examples too?

src/CAinputs.hpp
```diff
@@ -26,6 +26,7 @@ struct Inputs {
 
     std::string simulation_type = "", material_filename = "", grain_orientation_file = "";
     unsigned long rng_seed = 0.0;
+    int build_increment_outer = 5000;
```
streeve, line 29: This variable name will confuse me - frequency instead of increment would help

streeve (approved):

Issue #22, opened 2021-10-05
Use of Kokkos/device for initialization routines


MattRolchigo:
Currently, the code does the majority of the initialization on the host, copying over data to the device after initialization is complete. Much of this outside of file reading can be converted into kokkos and parallelized, speeding up code initialization and avoiding the need for allocation of a good portion of memory on the host (would not need mirror views on host and device - could initialize them on the device)
Necessary step will be converting to Kokkos::Random for some init routines

streeve:
I'm fine to close this @MattRolchigo - are any others enough on device that you would check them off (isn't tempinit?)?

PR #336, opened 2024-04-30
Ability to initialize simulations from vtk grain ID data

MattRolchigo:
Removed the deprecated ability to read the old substrate grain ID data format and replaced it with the ability to read grain ID data from a vtk file

### DIFF
--- a/src/CAparsefiles.cpp
+++ b/src/CAparsefiles.cpp
@@ -170,3 +170,24 @@ std::size_t checkForHeaderValues(std::string header_line) {
     }
     return header_size;
 }
+
+// Read and discard "n_lines" lines of data from the file
+void skipLines(std::ifstream &input_data_stream, const int n_lines) {
+    std::string dummy_str;
+    for (int line = 0; line < n_lines; line++)
+        getline(input_data_stream, dummy_str);
+}
+
+// Read space-separated tuple from a vtk file header
+std::vector<std::string> readVTKTuple(std::ifstream &input_data_stream) {
+    std::vector<std::string> read_values(3);
+    std::string read_line;
+    getline(input_data_stream, read_line);
+    std::size_t first_separator = read_line.find(' ');
+    std::size_t second_separator = read_line.find(' ', first_separator + 1);
+    std::size_t third_separator = read_line.find(' ', second_separator + 1);
+    read_values[0] = read_line.substr(first_separator, second_separator - first_separator);
+    read_values[1] = read_line.substr(second_separator, third_separator - second_separator);
+    read_values[2] = read_line.substr(third_separator);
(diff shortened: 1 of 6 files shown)


streeve:
Looks like this should check off part of #22 (maybe more of those are already done too?)

src/CAparsefiles.cpp
```diff
@@ -170,3 +170,24 @@ std::size_t checkForHeaderValues(std::string header_line) {
...
+std::vector<std::string> readVTKTuple(std::ifstream &input_data_stream) {
+    std::vector<std::string> read_values(3);
+    std::string read_line;
+    getline(input_data_stream, read_line);
+    std::size_t first_separator = read_line.find(' ');
+    std::size_t second_separator = read_line.find(' ', first_separator + 1);
+    std::size_t third_separator = read_line.find(' ', second_separator + 1);
+    read_values[0] = read_line.substr(first_separator, second_separator - first_separator);
+    read_values[1] = read_line.substr(second_separator, third_separator - second_separator);
+    read_values[2] = read_line.substr(third_separator);
```
streeve, line 191:
```suggestion
    char separator = ' ';
    std::size_t line_size = read_line.size();
    for (std::size_t n = 0; n < line_size - 1; n++) {
        std::size_t pos = read_line.find(separator);
        read_values[n] = read_line.substr(0, pos);
        read_line = read_line.substr(pos + 1, std::string::npos);
    }
```
Trying to make it more readable and reusable

MattRolchigo: Now that I think about it, we already have `splitString` that can do this exact thing...

src/CAparsefiles.cpp
```diff
@@ -82,6 +82,35 @@ void splitString(const std::string line, std::vector<std::string> &parsed_line,
...
+    getline(input_data_stream, line);
+
+    // Make sure the right number of values are present on the line - one more than the number of separators
+    std::size_t actual_num_values = std::count(line.begin(), line.end(), separator) + 1;
+    if (expected_num_values != actual_num_values) {
+        std::string error = "Error: Expected " + std::to_string(expected_num_values) +
+                            " values while reading file; but " + std::to_string(actual_num_values) + " were found";
+        throw std::runtime_error(error);
+    }
+    // Separate the line into its components, now that the number of values has been checked
```
streeve, line 100: Can't you just call `splitString`?

MattRolchigo: You mean call `void splitString` from within `std::vector<std::string splitString`?

streeve: Yeah that's what I meant

streeve (approved):

PR #342, opened 2024-05-15
Replaced layer_time_temp_history with liquidus_time and cooling_rate views of the appropriate types

### LINKED ISSUE #219: Exact storage of time at which cells go above/below liquidus
`LayerTimeTempHistory` currently stores the time steps at which the cells go above/below the liquidus as floats, which are later converted to integers for comparison with the current time step. This creates potential issues with large integers that cannot be successfully converted to float and back to int again. This also increases the small risk of a cell going above and below the liquidus on the same time step, which should not be allowed (these should be at least 1 time step apart, even if the difference between them is less than the time step)

MattRolchigo:
Closes #219  - also ensures that cells do not go above and then back below the liquidus on the same time step

### DIFF
--- a/src/CAnucleation.hpp
+++ b/src/CAnucleation.hpp
@@ -94,8 +94,8 @@ struct Nucleation {
             Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), temperature.max_solidification_events);
         auto number_of_solidification_events_host =
             Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), temperature.number_of_solidification_events);
-        auto layer_time_temp_history_host =
-            Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), temperature.layer_time_temp_history);
+        auto liquidus_time_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), temperature.liquidus_time);
+        auto cooling_rate_host = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace(), temperature.cooling_rate);
 
         // Use new RNG seed for each layer
         std::mt19937_64 generator(rng_seed + static_cast<unsigned long>(layernumber));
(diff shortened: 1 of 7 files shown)


src/CAtemperature.hpp
```diff
@@ -467,20 +472,20 @@ struct Temperature {
...
                 if (tot_dist <= spot_radius) {
                     // Melt time step (first time step)
-                    layer_time_temp_history_local(index, 0, 0) = 1;
-                    // Liquidus time step (related to distance to spot edge)
-                    layer_time_temp_history_local(index, 0, 1) =
-                        Kokkos::round((spot_radius - tot_dist) / isotherm_velocity) + 1;
+                    liquidus_time_local(index, 0, 0) = 1;
+                    // Liquidus time step (related to distance to spot edge). Must be at least 1 time step after melt
+                    // time step
+                    liquidus_time_local(index, 0, 1) = Kokkos::round((spot_radius - tot_dist) / isotherm_velocity) + 2;
```
streeve, line 478: Not sure what the original +1 was vs the new +1 after melt

src/CAnucleation.hpp
```diff
@@ -104,7 +104,7 @@ struct Nucleation {
         std::uniform_real_distribution<double> y_dist(-0.49999, grid.ny - 0.5);
         std::uniform_real_distribution<double> z_dist(-0.49999, grid.nz_layer - 0.5);
         // Gaussian distribution of nucleation undercooling
-        std::normal_distribution<float> g_distribution(_inputs.dtn, _inputs.dtsigma);
+        std::normal_distribution<double> g_distribution(_inputs.dtn, _inputs.dtsigma);
```
streeve, line 107: Not sure why this is needed

MattRolchigo: Because of the fact that some integers were not able to be converted to floats and back again, I replaced the use of `float` with `double` for the calculation of nucleation times and changed the std::normal_distributions (also used in the calculation) to `double` to match

src/CAtemperature.hpp
MattRolchigo: Just making sure that `liquidus_time_local(index, 0, 1)` is never equal to `liquidus_time_local(index, 0, 0)`

streeve (approved):

PR #344, opened 2024-05-21
Formatting of json string for output SubstrateFilename

MattRolchigo:


streeve (changes requested):
Test?

MattRolchigo:
There's no great way to test this specific changeset other than running ExaCA with a substrate file (which we don't have in the repository) and parsing the json file after the run. There's so many possible variables listed in the json output depending on problem type specifics that you'd need a large number of tests for every case. But I did test this on Frontier for another problem just now and it's working after the latest commit (previously seg faulted trying to read the improperly formatted json log during `grain_analysis` execution)

streeve:
> There's no great way to test this specific changeset other than running ExaCA with a substrate file (which we don't have in the repository) and parsing the json file after the run. There's so many possible variables listed in the json output depending on problem type specifics that you'd need a large number of tests for every case. But I did test this on Frontier for another problem just now and it's working after the latest commit (previously seg faulted trying to read the improperly formatted json log during `grain_analysis` execution)

Could you add an input test that just reads back in the created log file?

MattRolchigo:
What if I just append `testInputs` with the writing of the json log (right now it just ends after making sure the inputs were correctly read) and ensuring that it can be successfully read back in?

streeve (approved):

PR #346, opened 2024-05-21
Elimination of redundant memory access and calculations in cell capture

MattRolchigo:
Renamed variables and added comments to make cell capture subroutine more descriptive, and removed/optimized calculation of certain intermediate quantities. Up to a 5% performance improvement by reducing the amount of memory access (particularly for `diagonal_length` and `cell_type`)

### DIFF
--- a/src/CAupdate.hpp
+++ b/src/CAupdate.hpp
@@ -331,27 +331,45 @@ void cellCapture(const int, const int np, const Grid &grid, const InterfacialRes
                                 const float y2 = triangle_y[(triangle_index + 2) % 3];
                                 const float z2 = triangle_z[(triangle_index + 2) % 3];
 
-                                const float d1 = Kokkos::hypot(xp - x2, yp - y2, zp - z2);
-                                const float d2 = Kokkos::hypot(xc - x2, yc - y2, zc - z2);
-                                const float d3 = Kokkos::hypot(xp - x1, yp - y1, zp - z1);
-                                const float d4 = Kokkos::hypot(xc - x1, yc - y1, zc - z1);
+                                // Distance between the nearest corner of the capturing face (xc,yc,zc) and the other
+                                // two corners (should theoretically be the same, but may be slightly different due to
+                                // floating point errors) Previously d4
+                                const float dist_first_corner = Kokkos::hypot(xc - x1, yc - y1, zc - z1);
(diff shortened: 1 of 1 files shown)


src/CAupdate.hpp
```diff
@@ -313,68 +312,80 @@ void cellCapture(const int, const int np, const Grid &grid, const InterfacialRes
...
+                                    proj_next_nearest_corner_edge_2 = dist_second_corner - proj_nearest_corner_edge_2;
                                 }
-                                float l_12 = 0.5 * (Kokkos::fmin(i_1, Kokkos::sqrt(3.0f)) +
-                                                    Kokkos::fmin(i_2, Kokkos::sqrt(3.0f)));
-                                float l_13 = 0.5 * (Kokkos::fmin(j_1, Kokkos::sqrt(3.0f)) +
-                                                    Kokkos::fmin(j_2, Kokkos::sqrt(3.0f)));
+
+                                // Truncate the lengths at sqrt(3) for a max initial size of an octahedron
+                                const float l_12 =
+                                    0.5 * (Kokkos::fmin(proj_nearest_corner_edge_1, Kokkos::sqrt(3.0f)) +
```
streeve, line 368: Could make `sqrt(3)` a class constant or cache it above

streeve (commented):
All seems reasonable, but also very easy place to introduce a bug. Any thoughts (new test, throw caution to the wind, etc.)?

MattRolchigo:
Luckily we do have a test for this - CAupdate would've failed for the single grain or the directional solidification problem if we messed up (which I did do at one point), but cell capture seems to be working at intended based on the number of time steps it takes to complete the small example problems. I'll definitely want to test a couple large problems on GPU to make sure no weird floating point edge case got introduced though

src/CAupdate.hpp
MattRolchigo: I thought about doing that, I've seen that done in some older CPU codes before. I figured the memory access of the constant vs the time to compute sqrt(3) would be pretty comparable, but worth checking

MattRolchigo: It doesn't seem to have made a difference for performance either way, so I'll just leave it as is for simplicity

streeve (approved):

PR #351, opened 2024-05-22
Make sure all grid inputs for json write/read test are initialized

MattRolchigo:
Bug fix introduced when fixing bug in #344


streeve (approved):
How does the bug manifest and why did the tests pass?

MattRolchigo:
Since these member variables from `grid` are uninitialized, sometimes we get lucky and they get assigned a random float value that gets written into the json log and read back in. Sometimes they become nan and we get a parsing error when reading the json data back in (since the code expects a float number). I realized this back in #344 and set most of them appropriately, but I forgot about a few of them. I realized this only when this test randomly failed while testing something unrelated

Issue #350, opened 2024-05-22
Translation of input temperature data


MattRolchigo:
For problem types `FromFinch` and `FromFile`, we should have the ability to create spatially larger solidification domains by translating and offsetting input temperature data in space and time

PR #353, opened 2024-05-29
Translation/mirroring of Finch data for coupled runs

MattRolchigo:
Ability to duplicate Finch data by translating and optionally mirroring data to form larger scan areas. Will be useful to simulate unidirectional or bidirectional scan solidification using a single line scan dataset from Finch

Partially addresses #350 , adding analogous capabilities for the `FromFile` problem type will be done in a future changeset


src/CAinputs.hpp
```diff
@@ -157,6 +157,38 @@ struct Inputs {
                 // defaults to false
                 if (input_data["TemperatureData"].contains("TrimUnmeltedRegion"))
                     temperature.trim_unmelted_region = input_data["TemperatureData"]["TrimUnmeltedRegion"];
+                temperature.number_of_copies = input_data["TemperatureData"]["TranslationCount"];
+                // Include the original in the total number of copies (number of translations + 1)
+                temperature.number_of_copies++;
```
streeve, line 162: Maybe worth including in the docs - kind of implied by "default to 0"

streeve (approved):

examples/README.md
```diff
@@ -88,6 +88,11 @@ The .json files in the examples subdirectory are provided on the command line to
...
 |TemperatureFiles | FromFile             | List of files corresponding to each layer's temperature data, in the form ["filename1.csv","filename2.csv",...]. If the number of entries is less than NumberOfLayers, the list is repeated. Note that if the Z coordinate of the top surface for each data set has the layer offset applied, LayerOffset in the "Domain" section of the input file should be set to 0, to avoid offsetting the layers twice.
 |InitUndercooling | Directional, SingleGrain       | For SingleGrain, this is the undercooling at the location of the seeded grain. For problem type Directional, this is an optional argument (defaulting to zero) for the initial undercooling at the domain's bottom surface
+|TranslationCount       | FromFinch             | Number of translations of the Finch temperature data in the direction specified by `OffsetDirection`. Defaults to 0
```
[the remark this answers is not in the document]
MattRolchigo: Yes (and I'll need to for some future weak scaling/Toucan comparison stuff), but there's a bunch of additional code needed to do this with `FromFile`, so I was going to set that up as a separate pull request. I'm currently testing that on Frontier

src/CAinputs.hpp
MattRolchigo: So how would you suggest editing the current statement in the README: "Number of translations of the Finch temperature data in the direction specified by `OffsetDirection`. Defaults to 0"? Something about the fact that 0 still means we're using the temperature data, just that we aren't making any copies of it?

streeve: "Defaults to 0 (does not copy or translate data)." ?

PR #354, opened 2024-05-30
Ability to trim ExaCA domain based on the bounds of the Finch solidification data

### LINKED ISSUE #348: Finch-ExaCA solidification domain size
ExaCA uses the Finch simulation domain during coupled runs, which includes regions that do not undergo melting/solidification. The ExaCA domain should be trimmed to only bound the region that goes above the liquidus

MattRolchigo:
Previously, ExaCA used the same domain bounds as Finch, leading to unnecessary memory allocation by ExaCA as much of the Finch domain does not undergo melting or solidification.

Closes #348

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -116,8 +116,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: ORNL-MDF/Finch
-          # This is post-release 0.1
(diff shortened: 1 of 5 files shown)


streeve:
Need to update the CI file to use the most recent Finch commit hash

streeve (approved):

PR #355, opened 2024-05-31
Use exact bounds and times to set nucleation event locations and undercoolings

MattRolchigo:
Minor change to ensure the physical XYZ domain bounds are used to set the number of nuclei (not just the number of cells) and that nucleation events are ordered in time using exact time values rather than time step values. Also some minor variable renaming and unit test fixup to match the changes made


MattRolchigo:
The result for the SmallDirS unit test had to be updated as the nucleation events now occur in slightly different locations, leading to a slightly different final fraction of the domain filled by nucleated grains

streeve (approved):
Looks good

PR #360, opened 2024-06-14
Use physical Z layer bounds to set nucleation volume in placeNuclei and associated unit test

MattRolchigo:
Fixup of #355

### DIFF
--- a/src/CAnucleation.hpp
+++ b/src/CAnucleation.hpp
@@ -99,7 +99,8 @@ struct Nucleation {
         // in meters
         std::uniform_real_distribution<double> nucleation_site_dist_x(grid.x_min, grid.x_max);
         std::uniform_real_distribution<double> nucleation_site_dist_y(grid.y_min, grid.y_max);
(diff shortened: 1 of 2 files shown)


src/CAnucleation.hpp
```diff
@@ -142,6 +143,8 @@ struct Nucleation {
                 nuclei_x(n_event) = Kokkos::round((nuclei_x_unrounded - grid.x_min) / grid.deltax);
                 nuclei_y(n_event) = Kokkos::round((nuclei_y_unrounded - grid.y_min) / grid.deltax);
                 nuclei_z(n_event) = Kokkos::round((nuclei_z_unrounded - grid.z_min_layer[layernumber]) / grid.deltax);
+                std::cout << "Nuclei has " << nuclei_x(n_event) << ", " << nuclei_y(n_event) << ", "
```
streeve, line 146: Printing on all ranks?

MattRolchigo: Definitely was leftover from when I was debugging the unit test and wanted it to print out the locations of the nuclei

streeve (approved):

PR #363, opened 2024-06-28
Remove redundant simulation type check

MattRolchigo:
Not sure how this got into the intermediate output subroutine to begin with, the simulation type only needs to be checked during the input read since it doesn't change at all during a run

### DIFF
--- a/src/CAgrid.hpp
+++ b/src/CAgrid.hpp
@@ -126,9 +126,6 @@ struct Grid {
         , _inputs(inputs)
         , _t_inputs(t_inputs) {
 
-        // Check for valid simulation type.
-        validSimulationType(simulation_type);
-
         // Copy from inputs structs
         deltax = _inputs.deltax;
         layer_height = _inputs.layer_height;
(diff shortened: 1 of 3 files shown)


streeve (commented):
I put it anywhere we passe the type, generally in case you weren't using the inputs directly. Is it worth it? I don't know, but there are two other places aside from the initial check during input parsing:

src/CAgrid.hpp:        validSimulationType(simulation_type);
src/CAinputs.hpp:        validSimulationType(simulation_type);
src/CAtemperature.hpp:        validSimulationType(simulation_type);

MattRolchigo:
It's not worth it - simulation type does not change once it is set in the constructor, if it's an invalid simulation type that should be caught during initialization but it doesn't need to be checked every time

streeve (approved):
If there's ever a strange error because of a mistaken simulation type I will gloat

MattRolchigo:
Fair enough, I felt good about this because the only times that we pass the problem type to a function, its the only input of type std::string so it won't get mixed up with another input for now

PR #364, opened 2024-07-12
Code and example problem cleaup

MattRolchigo:
Minor changes to comments to ensure they are up to date with the current code, also changed a couple of variable names for consistency with the current capitalization scheme, and changed a few std::cout messages.

Only changes to the actual code functionality are the two commits that are separated out - a portion of the powder layer init is skipped if the number of powder layer cells to be assigned grain ID values is 0 (this also avoids printing a confusing std::cout message about assigning cells non-existent powder layer grain IDs, and the translation increment in the two relevant example files was updated to ensure the temperature data actually overlaps.

### DIFF
--- a/src/CAcelldata.hpp
+++ b/src/CAcelldata.hpp
@@ -578,53 +578,57 @@ struct CellData {
         int powder_layer_cells = grid.nx * grid.ny * powder_layer_height;
         int powder_layer_assigned_cells =
             Kokkos::round(static_cast<double>(powder_layer_cells) * _inputs.powder_active_fraction);
-        // Associate powder grain IDs with CA cells in the powder layer
         MPI_Barrier(MPI_COMM_WORLD);
         if (id == 0)
             std::cout << "Initializing powder layer for Z = " << powder_bottom_z << " through " << powder_top_z - 1
                       << " (" << grid.nx * grid.ny * powder_layer_height << " cells): powder layer has "
                       << powder_layer_assigned_cells << " cells assigned new grain ID values" << std::endl;
-        std::vector<int> powder_grain_ids(powder_layer_cells, 0);
-        for (int n = 0; n < powder_layer_assigned_cells; n++) {
-            powder_grain_ids[n] = n + next_layer_first_epitaxial_grain_id; // assigned a nonzero GrainID
-        }
-        std::shuffle(powder_grain_ids.begin(), powder_grain_ids.end(), gen);
-        // Wrap powder layer GrainIDs into an unmanaged view, then copy to the device
-        view_type_int_unmanaged powder_grain_ids_host(powder_grain_ids.data(), powder_layer_cells);
-        auto powder_grain_ids_device = Kokkos::create_mirror_view_and_copy(memory_space(), powder_grain_ids_host);
 
-        int powder_start = grid.nx * grid.ny * powder_bottom_z;
-        if (id == 0)
-            std::cout << "Powder layer grain ID values range from " << next_layer_first_epitaxial_grain_id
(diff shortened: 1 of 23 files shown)


src/CAcelldata.hpp
```diff
@@ -88,11 +88,12 @@ struct CellData {
                       << " initialized at X = " << grain_location_x << ", Y = " << grain_location_y
                       << ", Z = " << grain_location_z << std::endl;
 
+        MPI_Barrier(MPI_COMM_WORLD);
```
streeve, line 91: Lots of extra barriers.. I assume this is extra-precautionary

src/CAgrid.hpp
```diff
@@ -318,16 +322,10 @@ struct Grid {
         return xyz_min_max;
     }
 
-    // For simulation type R, obtain the physical XYZ bounds of the domain by reading temperature data files and parsing
-    // the coordinates. Previously in CAinitialize.cpp
+    // For simulation type FromFile, obtain the physical XYZ bounds of the domain by reading temperature data files and
+    // parsing the coordinates (does not store temperature data points). Previously in CAinitialize.cpp
```
streeve, line 326: Do we care that it used to be in a different file?

src/CAtemperature.hpp
```diff
@@ -98,7 +98,7 @@ struct Temperature {
         getCurrentLayerUndercooling(grid.layer_range);
     }
 
-    // Constructor using in-memory temperature data from external source.
+    // Constructor using in-memory temperature data from external source (input_temperature_data
```
streeve, line 101: )

src/CAtemperature.hpp
```diff
@@ -951,9 +947,9 @@ struct Temperature {
     }
 
     // Extract either the last time step that all points undergo melting in the layer or the last time they cools below
-    // the liquidus from layer_time_temp_history (corresponds to solidification event number `NumSolidificationEvents-1`
-    // for the cell) (can't just use subview here since NumSolidificationEvents is different for each cell) If the cell
-    // does not undergo solidification, either print -1 or the specified default value
+    // the liquidus from liquidus_time (corresponds to solidification event number `number_of_solidification_events-1`
```
streeve, line 950: May as well remove these single quotes

src/CAcelldata.hpp
MattRolchigo: Yeah I tried to be consistent and have cout and a MPI_Barrier at the end of each constructor/init routine (it was already in most of them) so that the location of failures in the future can be identified more easily. Only one without the barrier is orientations, since its used in the grain analysis (which is serial and will fail with a call to MPI_Barrier without MPI initialized)

src/CAgrid.hpp
MattRolchigo: Honestly, probably not (that was mostly in there for my sake to keep things straight during the big code reorganization), but I figured I'd wait until the next release to remove the references to the old file

src/CAinputs.hpp
```diff
@@ -772,6 +773,9 @@ struct Inputs {
             throw std::runtime_error("Error: Unrecognized functional form for interfacial response function, currently "
                                      "supported options are quadratic, cubic, and exponential");
         irf.freezing_range = data["freezing_range"];
+        MPI_Barrier(MPI_COMM_WORLD);
+        if (id == 0)
```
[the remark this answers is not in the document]
MattRolchigo: I think I got all of them in this last edit (in CAinputs.hpp)

src/CAcelldata.hpp
streeve: Seems heavy handed, but it is more consistent now and only in init (as far as I can tell)

streeve (approved):

PR #369, opened 2024-08-06
Remove unused grain analysis LayerID read and functionality

MattRolchigo:
* Removed unused function in `GAutils.hpp`
* Fixed bug in reading/ignoring ASCII fields (previously didn't affect analysis due to the way it was written)
* LayerID is now an optional vtk field in the ExaCA output, if not given, it is noted that certain information cannot be gained from the output data (we don't know whether or not a cell underwent melting/solidification) but the analysis is still performed on all cells with non-zero GrainID values
* Added unit test for vtk file reading with various permutations of fields given and formatting


analysis/src/GAutils.hpp
```diff
@@ -168,12 +124,21 @@ void initializeData(std::string microstructure_file, int nx, int ny, int nz, Vie
...
                 }
                 std::cout << "Data field " << read_fieldname << " read" << std::endl;
             }
         }
     }
     input_data_stream.close();
+    if (!found_grain_id)
+        throw std::runtime_error("Error: analysis requires the GrainID field to be present in the vtk file");
+    if (!found_layer_id) {
+        std::cout << "Note: LayerID not present in vtk data, analysis will not differentiate between cells that did "
```
streeve, line 137: Seems like a warning to me

MattRolchigo: I don't know, I feel like normally I used warnings for something that's heavily discouraged or will be deprecated. A dataset without LayerID present will only lead to one minor analysis option being excluded and slightly different orientation histogram data in the specific case where there's a nonzero powder density and cells that do not go above/below the liquidus.

streeve (approved):

analysis/src/GAutils.hpp
streeve: Fair point

PR #370, opened 2024-08-12
Fix device print usage

streeve:
Only print with Kokkos::printf (introduced in version 4.2)

Previously broke with SYCL backend

### DIFF
--- a/src/CAinterface.hpp
+++ b/src/CAinterface.hpp
@@ -255,9 +255,8 @@ struct Interface {
     }
 
     // Load data (grain_id, octahedron_center, diagonal_length) into ghost nodes if the given coord_y is associated with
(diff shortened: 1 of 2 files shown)


MattRolchigo:
@streeve is there a way to check the CI for exactly which temperature unit test is failing? I've seen intermittent failure of the unit test but never twice in a row like this

streeve:
> @streeve is there a way to check the CI for exactly which temperature unit test is failing? I've seen intermittent failure of the unit test but never twice in a row like this

Ran fine this time.. we would either need to run the tests with verbose output all the time or add the option to inject it when we want (and hope it fails that time)

MattRolchigo (approved):

PR #372, opened 2024-08-13
Shared Voronoi functionality between baseplate and powder init

MattRolchigo:
An edge case was previously possible in ExaCA where, due to the fact that the powder layer initialization left some cells with a "zero" GrainID, simulations of FromFile and FromFinch problems could get locked in a loop where no valid grain could claim the liquid cells in some melt pools for many time steps, often leading to simulation timeout. It also lead to unnatural grain shapes as grains stretch across multiple melt pools.

This changeset uses Voronoi helper functions to initialize both the baseplate (which previously used a Voronoi tessellation to assigned GrainIDs to cells) and each powder layer, so that no cells have a GrainID of 0 following initialization. The PowderDensity input is replaced with a MeanPowderGrainSize input, and for consistency, the MeanSize input is renamed MeanBaseplateGrainSize. Minor changes to the std::cout statements are also made to log which GrainID values are being used for which parts of the simulation (baseplate, powder layers, nucleated grains)

### DIFF
--- a/src/CAinputs.hpp
+++ b/src/CAinputs.hpp
@@ -314,7 +314,7 @@ struct Inputs {
                 ((input_data["Substrate"].contains("MeanSize")) ||
                  (input_data["Substrate"].contains("MeanBaseplateGrainSize"))))
                 throw std::runtime_error(
-                    "Error: only one of substrate grain size and substrate structure filename should "
+                    "Error: only one of baseplate grain size and substrate structure filename should "
                     "be provided in the input file");
             else if (input_data["Substrate"].contains("SubstrateFilename")) {
                 substrate.substrate_filename = input_data["Substrate"]["SubstrateFilename"];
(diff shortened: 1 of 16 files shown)


src/CAinputs.hpp
```diff
@@ -724,7 +747,7 @@ struct Inputs {
                     exaca_log << "       \"SubstrateFilename\": "
                               << "\"" << substrate.substrate_filename << "\"" << std::endl;
                 else
-                    exaca_log << "       \"MeanSize\": " << substrate.substrate_grain_spacing << std::endl;
```
MattRolchigo, line 727: Left this as "MeanSize" for now but since this isn't used in any post-processing, it is probably safe to just rename it and add the powder grain size to the log

unit_test/tstInputs.hpp
```diff
@@ -43,6 +43,7 @@ void writeTestData(std::string input_filename, int print_version) {
                    << std::endl;
     test_data_file << "   }," << std::endl;
     test_data_file << "   \"Substrate\": {" << std::endl;
+    // Note: old input used here to check that the expected mean powder grain spacing ends up correct
     test_data_file << "      \"SubstrateFilename\": \"DummySubstrate.vtk\"," << std::endl;
     test_data_file << "      \"PowderDensity\": 1000," << std::endl;
```
MattRolchigo, line 48: This was left as the deprecated input on purpose to check and make sure the conversion to powder grain size is correct, but this should be changed when the deprecated input is removed

streeve (approved):

PR #376, opened 2024-09-13
Avoid erroneously printing buffer load failure message in cell capture

MattRolchigo:
Fixup of #370 - the print should only occur in the refill buffer routine


streeve (approved):
Is a bug-ception always me "fixing" a mistake and then you actually fixing it?

MattRolchigo:
@streeve no, there's multiple instances of me making a dumb mistake and then finding it some time between 1 minute and 1 month after I merged something that was "just a minor change"

PR #377, opened 2024-09-20
Update dev version number

streeve:


MattRolchigo (approved):

MattRolchigo:
Should this be added to the changelog with a note that v2.0 erroneously printed the wrong version number?

streeve:
> Should this be added to the changelog with a note that v2.0 erroneously printed the wrong version number?

I'll include that in the 2.0.1 notes

PR #388, opened 2025-02-05
Optional two phase solidification with two interface response functions

MattRolchigo:
Before merging:

- [x] Example problem (AM preferred) to show two phase capability using the new Material file SS316_FA.json
- [x] Unit tests for phase selection and growth capabilities
- [x] README updates

Update 4/25: Slowdown of around 2% from master branch to this one, not 10% as previously noted


src/CAinterfacialresponse.hpp
```diff
@@ -23,63 +23,85 @@ struct InterfacialResponseFunction {
...
+            _inputs.A[phase_num] *= static_cast<float>(deltat / deltax);
+            _inputs.C[phase_num] *= static_cast<float>(deltat / deltax);
         }
     }
 
     // Compute velocity from local undercooling.
     // functional form is assumed to be cubic if not explicitly given in input file
     KOKKOS_INLINE_FUNCTION
-    float compute(const float loc_u) const {
+    float compute(const float loc_u, const int phase_num = 0) const {
```
streeve, line 54: `assert( phase_num <= num_phases);`

examples/Materials/SS316L_FA.json
streeve, line 1: Probably worth naming with AusteniteFerrite for clarity

examples/Inp_SmallSpotMelt_Steel2.json
streeve, line 1: Steel_2phase?

streeve (commented):
Just small stuff

analysis/src/GArepresentativeregion.hpp
```diff
@@ -369,6 +373,24 @@ struct RepresentativeRegion {
         return grain_id_vector;
     }
 
+    // Get the list of Phase IDs associated with the representative region (same as getGrainIDVector except for return
```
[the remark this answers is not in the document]
MattRolchigo: I don't actually use this vector for anything and it turns out that the vector would be kind of useless (since it'll just be a bunch of 0s and 1s decoupled from the grains they're associated with), so I'm going to remove this function and modify the analysis routine a bit on the next commit

unit_test/tstInputs.hpp
```diff
@@ -140,11 +140,11 @@ void testInputs(int print_version) {
         EXPECT_DOUBLE_EQ(inputs.nucleation.dtsigma, 0.5);
         // Raw values read from material input file - not normalized (normalization occurs on local copies of these in
         // the IRF constructor)
-        EXPECT_FLOAT_EQ(inputs.irf.A, -0.00000010302);
-        EXPECT_FLOAT_EQ(inputs.irf.B, 0.00010533);
-        EXPECT_FLOAT_EQ(inputs.irf.C, 0.0022196);
-        EXPECT_FLOAT_EQ(inputs.irf.D, 0);
-        EXPECT_FLOAT_EQ(inputs.irf.freezing_range, 210);
+        EXPECT_FLOAT_EQ(inputs.irf.A[0], -0.00000010302);
```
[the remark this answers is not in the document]
MattRolchigo: This is already tested for in `testInterfacialResponse` - we read both the old `SS316L.json` and the new two phase file and check that the inputs were initialized correctly for each constant and each applicable phase. It's kind of redundant that we also do this check in the main input file parsing anyways

src/CAinputdata.hpp
```diff
@@ -48,17 +48,23 @@ struct NucleationInputs {
...
         cubic = 0,
         quadratic = 1,
         power = 2,
+        exponential = 3,
     };
-    int function = cubic;
+    enum Transformationtypes { none = 0, solidification = 1 };
+    int function[2] = {cubic, cubic};
+    int transformation = none;
+    bool velocity_cap[2] = {false, false};
```
streeve, line 67: Should this be documented more somewhere?

src/CAinputdata.hpp
```diff
@@ -48,17 +48,23 @@ struct NucleationInputs {
...
 
 struct InterfacialResponseInputs {
-    float freezing_range;
-    float A;
-    float B;
-    float C;
-    float D = 0.0;
+    int num_phases = 1;
+    std::vector<std::string> phase_names;
+    float freezing_range[2];
```
streeve, line 53: Just pointing out again (probably unhelpfully) that this is a bit cumbersome for single phase and for extension to three phase

MattRolchigo: I can't really think of a better way, I guess you could have multiple InterfaceResponseFunction objects which was my original plan but I remember running into an issue with that. At the very least, doing it this way involves changing less of the rest of the code and I can't really think of a situation where more than 2 solid phases form at the grain scale that we'd be interested in modeling

streeve (approved):
Still approved, still not the most thorough review

PR #390, opened 2025-02-21
Option to specify location of baseplate bottom

MattRolchigo:
For "FromFile" or "FromFinch" type simulations, ExaCA currently sets the bottom of the baseplate to be the smallest Z coordinate that contains temperature data. For one-to-one comparisons across time-temperature history datasets where that smallest Z coordinate varies, there is now an input option `[Substrate][BaseplateBottomZ]`to set a fixed Z coordinate to be the bottom of the baseplate. The default behavior is unchanged.


src/CAgrid.hpp
```diff
@@ -121,16 +124,28 @@ struct Grid {
             y_min = _t_inputs.temperature_y_bounds[0];
             y_max = _t_inputs.temperature_y_bounds[1];
         }
+        if (_s_inputs.use_fixed_z_bounds) {
```
streeve, line 127: Seems easier to not store this and do this check always even if you know it won't be below because you extracted from the data (just my preference)

src/CAgrid.hpp
```diff
@@ -121,16 +124,28 @@ struct Grid {
...
+            // Check to ensure that the specified lower Z bound is smaller than the smallest Z with temperature data
+            // (otherwise throw an error)
+            if (z_min < _s_inputs.baseplate_bottom_z)
+                throw std::runtime_error("Error: temperature data extends below the input value for BaseplateBottomZ");
+            if (id == 0)
+                std::cout << "Overriding the lower Z bound given by the temperature data (" << z_min
```
streeve, line 133: Seems like a strange message - isn't that exactly the intent of using that option?

src/CAgrid.hpp
MattRolchigo: I did this for consistency with how we handle manually set X and Y domain bounds - the override function already checks for these bounds in the same way, so adding a check for Z as well made the most sense to me

src/CAgrid.hpp
MattRolchigo: It is the intent of using the option - and it's the same message we print when X and Y bounds are manually specified. Since the X, Y, and Z bounds are printed to the log without indication of whether they were from the data or manually specified, it makes sense to me that we'd print to the screen whether or not we manually supplied bounds. Kind of like we also echo other information about the inputs read.

streeve (approved):
I understand better now that I read those preceding lines. Slightly confusing this belongs to substrate instead, but I understand why

PR #394, opened 2025-05-06
Default to number of layers = 1 for single layer problem types

MattRolchigo:
Fixed issue where running analysis on `Spot` or `SingleGrain` problem type output would throw an error due to a lack of a default for number of layers. Also added a warning on deprecated problem type `S`


streeve (approved):
Looks good. When are we removing these deprecated features?

PR #395, opened 2025-05-06
Correct name of analysis input option printAvgStats

MattRolchigo:
The incorrect name `printStats` was erroneously used in place of the analysis input argument `printAvgStats`


streeve:
Looks like a real failure

streeve (approved):

PR #398, opened 2025-06-04
In-memory Finch-ExaCA coupling with multiple Finch simulations

MattRolchigo:
To be merged concurrently with corresponding pull request in the Finch repository


unit_test/test_harness_m.cmake
streeve, line 1: Remove?

streeve:
Blocked by https://github.com/ORNL-MDF/Finch/pull/28

.github/workflows/CI.yml
streeve: Should this be updated?

examples/README.md
```diff
@@ -14,9 +14,9 @@ ExaCA currently can model five types of problems:
...
+The .json files in the examples subdirectory are provided on the command line to let ExaCA know which problem is being simulated. The json files contain different input sections, with certain inputs only used for certain problem types. For optional inputs, the default value used is noted. There are two types of input files for ExaCA: "singular" (ExaCA) input files and "combined" (Finch-ExaCA). For combined input files, the top level input sections are `Finch` and `ExaCA`, with subsections `SimulationType`, `MaterialFileName`, `GrainOrientationsFile`, `RandomSeed`, `Domain`, `Nucleation`, `TemperatureData`, `Substrate`, and `Printing` in the ExaCA portion of the file and subsections as given in the Finch README for the Finch portion of the file. Note that for the Finch section of these files, inputs that vary on different layers (for multilayer simulations) can be given in a `layers` list; see `examples/Inp_FinchVariedLayers.json` for an example of this. For singular input files, the aforemenioned ExaCA subsections are top level inputs. 
 
-## Top level inputs
+## Top level inputs (for singular type input files) or second level inputs (for combined input files)
```
streeve:
```suggestion
## Top level inputs (individual input files) or ExaCA level inputs (combined input files)
```

.github/workflows/CI.yml
MattRolchigo: It should work either way due to the backwards compatibility (the fact that the test passed is a good sign), but yes I should change this to account for the preferred method of supplying Finch-ExaCA input files (1 combined input file instead of two)

examples/Inp_FinchVariedLayers.json
```diff
@@ -0,0 +1,87 @@
...
+        },
+        "sampling": {
+          "type": "solidification_data",
+          "format": "default",
+          "directory_name": "single_line_small_exaca"
+        },
+        "layers": [
+            {
+                "source": {
+                    "absorption": 0.4,
```
streeve, line 33: How realistic is this case where the absorption changes layer by layer?

streeve (approved):

examples/Inp_FinchVariedLayers.json
MattRolchigo: Not particularly realistic, but this is essentially a proxy for changing the power (which we don't change since we use the same scan path each layer)

PR #409, opened 2025-08-29
Increment nucleation counter as necessary when time steps are skipped

MattRolchigo:
Fixes edge case bug for large mean nucleation undercooling and standard deviations where a candidate nucleation event can be assigned to a time step that is "skipped" by ExaCA due to no liquid cells existing on said time step - in turn this would lead to the nucleation event counter being unable to advance.

### DIFF
--- a/src/CAnucleation.hpp
+++ b/src/CAnucleation.hpp
@@ -315,6 +315,17 @@ struct Nucleation {
             }
         }
     }
+
+    // If the time step jumped from an older value to a newer value of "cycle", update nucleation_counter accordingly as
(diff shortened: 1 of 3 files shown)


src/CAnucleation.hpp
```diff
@@ -315,6 +315,17 @@ struct Nucleation {
             }
         }
     }
+
+    // If the time step jumped from an older value to a newer value of "cycle", update nucleation_counter accordingly as
```
streeve, line 319: Why is cycle sarcastic?

MattRolchigo: Because the variable name is dumb

PR #411, opened 2025-09-03
format: do not align trailing comments

streeve:
Noted in #409

### DIFF
--- a/src/CAinputs.hpp
+++ b/src/CAinputs.hpp
@@ -35,7 +35,7 @@ struct Inputs {
     std::string file_name;
 
     // Creates input struct with uninitialized/default values, used in unit tests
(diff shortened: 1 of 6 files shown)


src/CAinputs.hpp
```diff
@@ -35,7 +35,7 @@ struct Inputs {
     std::string file_name;
 
     // Creates input struct with uninitialized/default values, used in unit tests
-    Inputs() {};
+    Inputs(){};
```
streeve, line 38: This seems unrelated

MattRolchigo (approved):

src/CAinputs.hpp
streeve: Leaving as is

PR #418, opened 2025-09-30
Fix buffer unpack edge case where liquid cell data is unpacked into an existing liquid cell

MattRolchigo:
Previously, it was assumed that if a halo region cell is liquid and data from another MPI rank is received, that the received data must correspond to active cell data. This guards against an edge case where this is not true (i.e., attempting to unpack liquid cell data but the cell is already liquid) by checking that the unpacked octahedron diagonal length > 0.


streeve (approved):
So the comment already referred to this, but it got removed at some point?

MattRolchigo:
I don't know if the code ever had a check for the "non-zero diagonal length" mentioned in the comment, I think the erroneous assumption that "if the cell is liquid, we must be updating it with active cell data and therefore the new diagonal length will always be > 0" was made at some point, which works like 99% of the time

PR #430, opened 2026-03-11
feat: expand validation options for FromFile data headers

### LINKED ISSUE #429: feature request: more flexible header names for `FromFile` data
The current headers for ExaCA input files in the `FromFile` type of simulation (`x,y,z,tm,ts,cr`) are only partially self-descriptive, because they do not allow for unit specification in the header.

Allowing the header to include the base ExaCA units would be helpful in enabling the ExaCA input files to be better described.

For example, due to the existing pre-formatting before the header check (lowercase, remove spaces), adding "x(m)" as an acceptable alternative header for "x" would allow a few options: `x`, `X`, `x(m)`, `X(m)`, `x (m)`, and `X (m)`.

Alternate headers with units would be:

`"x(m)", "y(m)", "z(m)", "tm(s)", "tl(s)"|"ts(s)", and "r(k/s)"|"cr(k/s)"`

gknapp1:
Closes #429.

This allows data files used with FromFile to contain the units in the column names, which enables the data files to be more self-descriptive.

### DIFF
--- a/src/CAparsefiles.cpp
+++ b/src/CAparsefiles.cpp
@@ -161,7 +161,14 @@ std::size_t checkForHeaderValues(std::string header_line) {
     std::vector<std::string> header_values(header_size, "");
     splitString(header_line, header_values, header_size);
 
-    std::vector<std::vector<std::string>> expected_values = {{"x"}, {"y"}, {"z"}, {"tm"}, {"tl", "ts"}, {"r", "cr"}};
+    std::vector<std::vector<std::string>> expected_values = {
+        {"x",  "x(m)"},
+        {"y",  "y(m)"},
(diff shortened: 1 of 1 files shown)


MattRolchigo (approved):
This looks good to me - the only other thing I could think of is adding support for a capital K for kelvin as well in the units for cooling rate

gknapp1:
> This looks good to me - the only other thing I could think of is adding support for a capital K for kelvin as well in the units for cooling rate

The way I read it, the headers are already parsed with ::to_lower earlier in the file, so the expected value check should be all lowercase.

MattRolchigo:
Oh yeah you're right, I forgot Sam added that feature

PR #431, opened 2026-04-13
Added example for grain structure calibration

MattRolchigo:
Use of TASMANIAN and adaptive sampling to perform an ensemble of directional solidification calculations, plot output surrogate response surfaces, and calibrate the model inputs


utilities/TASMANIAN/DirSolCalibration.py
```diff
@@ -0,0 +1,368 @@
...
+    # Ensemble of simulations with varied nucleation density and thermal gradient
+    # Fixed isotherm velocity (m/s)
+    isotherm_velocity = 0.1
+
+    # Target mean grain area and mean grain extent in Z
+    target_ouputs = [100.0, 15.0]
+    # Tasmanian uniform grid
+    num_inputs = 2
+    num_outputs = 2
+    grid_res = 1
```
streeve: Why not throw these into another json file and read that in?

utilities/TASMANIAN/DirSolCalibration.py
```diff
@@ -0,0 +1,368 @@
...
+    plt.colorbar(im2)
+    ax2.contour(xs, ys, grain_extent_z_surrogate, colors='k', levels=[target_outputs[1]], linewidths=1.5)
+
+    ax2.set_xlabel("Thermal gradient (K/m)")
+    ax2.set_ylabel("Nucleation density ($m^{-3}$)")
+    ax2.set_title("Mean grain size in Z (µm)")
+    plt.savefig("surrogate_mean_grain_extent_z.png", dpi=300, bbox_inches='tight')
+    #plt.show()
+
+def plot_contours(grid, phys_bounds):
```
streeve:
```suggestion
def plotContours(grid, phys_bounds):
```
:)

utilities/TASMANIAN/DirSolCalibration.py
```diff
@@ -0,0 +1,368 @@
...
+    xs = phys_bounds[0][0] + (phys_bounds[1][0] - phys_bounds[0][0])*(x - np.min(XX))/(np.max(x) - np.min(x))
+    ys = phys_bounds[0][1] + (phys_bounds[1][1] - phys_bounds[0][1])*(y - np.min(YY))/(np.max(y) - np.min(y))
+
+    fig, ax = plt.subplots(ncols=1, nrows=1, constrained_layout=True)
+    ax.contour(xs, ys, mean_grain_area_surrogate, colors='k', levels=[0.68], linewidths=1.5)
+    ax.contour(xs, ys, grain_extent_z_surrogate, colors='k', levels=[exp_mean_grain_area], linewidths=1.5)
+    ax.set_xlabel("Nucleation density ($m^{-3}$)")
+    ax.set_ylabel("Mean nucleation undercooling (K)")
+    ax.set_title("Area fraction of as-solidified ferrite")
+    plt.savefig("/Users/65y/Desktop/AMMT_FY25/Calibration/Contours.png", dpi=300, bbox_inches='tight')
```
streeve:
```suggestion
    plt.savefig(folder_name + "Contours.png", dpi=300, bbox_inches='tight')
```

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: Mainly because this is just an example for SimCamp and I don't need this to be that customizable at the moment

gknapp1: Would it make more sense to put it in the `examples` folder? I would expect it to be more easily scriptable, either through command line arguments or JSON input, if it is in the `utilities` folder

utilities/TASMANIAN/DirSolCalibration.py
gknapp1, line 1: You should create a `requirements.txt` file for the Python dependencies (create a clean Python virtual environment, install all dependencies, and then run `pip freeze > requirements.txt`).

gknapp1, line 1: If this is intended to be a utility, then would definitely create an input file and probably update many of the hard-coded parameters to read from the input file.

If this is intended to be an example, then no glaring issues, but I haven't tested it yet.

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: The `examples` folder is for ExaCA input files specifically, while this script is for generation of ExaCA input files (it does put the generated ExaCA input files in `examples` though). I considered everything in `utilities` to be "inputs to something other than ExaCA, but used in tandem with ExaCA", so like Finch input files, pre- and post-processing scripts to generate ExaCA inputs files or analyze ExaCA output files, etc.

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: That's a good idea - I added a requirements file and a note in the README

utilities/TASMANIAN/DirSolCalibration.py
streeve: Maybe renaming `TestTasmanian` to describe the physics instead will help. Right now I find the new vs old tasmanian examples a bit unclear

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: This is intended as an example of how one could generate a series of input files using Tasmanian, but was placed in utilities since it isn't directly used by ExaCA

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: I removed `TestTasmanian` since it was an outdated example to begin with and the TASMANIAN usage demonstrated by it overlapped with the new example

README.md
```diff
@@ -258,11 +258,7 @@ The analysis executable, in addition to outputting grain statistics, can also ou
...
+Within the [utilities/](utilities/) directory, an example python script including automated input file generation, launching of an ensemble of ExaCA simulations and analysis of the simulation results, the plotting of surrogate model response surfaces for outputs of interest, and the calibration of the model input parameters to produce target outputs can be run using the python script [utilities/TASMANIAN/DirSolCalibration.py](utilities/TASMANIAN/DirSolCalibration.py). Python dependencies to run this script are given in [utilities/TASMANIAN/requirements.txt](utilities/TASMANIAN/requirements.txt). This example is based on the directional solidification test problem from the `examples` directory, where thermal gradient and heterogeneous nucleation density are the two inputs varied across the simulation ensemble members and mean grain area at the top surface and mean grain extent in the Z direction are the two outputs used for calibration.
```
streeve:
```suggestion
Within the [utilities/](utilities/) directory, an example python script for automating an ensemble of ExaCA simulations using TASMANIAN is provided. 
This includes input file generation, launching of an ensemble of ExaCA simulations, analysis of the simulation results, plotting of surrogate model response surfaces for outputs of interest, and the calibration of the model input parameters to match target outputs.
The python file [utilities/TASMANIAN/DirSolCalibration.py](utilities/TASMANIAN/DirSolCalibration.py) and python dependencies necessary [utilities/TASMANIAN/requirements.txt](utilities/TASMANIAN/requirements.txt) are available, together with dependent scripts.
This example is based on the directional solidification test problem from the `examples` directory, where thermal gradient and heterogeneous nucleation density are the two inputs varied across the simulation ensemble members and mean grain area at the top surface and mean grain extent in the Z direction are the two outputs used for calibration. 
```

utilities/TASMANIAN/DirSolCalibration.py
streeve: One last attempt - if you use `argparse` at least it's easy for users to change these

streeve (commented):
Will approve eventually if you refuse to make the inputs accessible outside the file

utilities/TASMANIAN/DirSolCalibration.py
MattRolchigo: I did end up adding an input file so that its easier to change these

streeve (approved):

PR #432, opened 2026-05-26
Avoid assigning color to non-existent phase in MTEX plotting utility

MattRolchigo:
For the edge case where two phases are given on the command line but no pixels are assigned to the second phase, this change avoids a Matlab error where the color attribute is assigned to the second phase of a single-phase EBSD object

### DIFF
--- a/utilities/MTEX/PlotIPFColoredSection.m
+++ b/utilities/MTEX/PlotIPFColoredSection.m
@@ -57,7 +57,8 @@ function [] = PlotIPFColoredSection(MTEXFile, Phaselist)
         figure(fig_counter);
         ebsd(pname1).color = [0 0 0];
         % only assign white color to second phase if ebsd object is multiphase
-        if (sum(ebsd.phase) ~= length(ebsd.phase))
+        second_phase_present = any(ebsd.phase == 2, 'all');
+        if (second_phase_present)
             ebsd(pname2).color = [1 1 1];
         end
         plot(ebsd);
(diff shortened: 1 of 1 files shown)


utilities/MTEX/PlotIPFColoredSection.m
```diff
@@ -56,7 +56,10 @@
     if (numphases == 2)
         figure(fig_counter);
         ebsd(pname1).color = [0 0 0];
-        ebsd(pname2).color = [1 1 1];
+        % only assign white color to second phase if ebsd object is multiphase
+        if (sum(ebsd.phase) ~= length(ebsd.phase))
```
streeve:
```suggestion
        if (sum(ebsd.phase) == 2)
```
Am I understanding this at all?

MattRolchigo: No, ebsd.phase accesses the phase ID of all of the (x,y) data, which is 1 for the default phase and 0 for the second phase. If no (x,y) locations were actually assigned to the second phase, the sum of the phase ID will equal the length of the phase ID vector, triggering this condition. It seems that if no (x,y) are present with a phase label "NameOfOtherPhase", you can't access `ebsd("NameOfOtherPhase").color`, which was triggering an error

streeve: How about `if (np.any(ebsd.phase == 0))`

streeve (approved):

PR #435, opened 2026-07-10
Added material file for AlSi10Mg

MattRolchigo:
Power law interface response function for solidification of AlSi10Mg alloy, with data from Mohebbi et al. Additive Manufacturing 36 (2020) 101726


colemanjs (approved):
Looks good

